Issue #20, opened 2016-01-20
Source code version mismatch


uded:
It seems like the current source code is no longer compatible with Java 5 level.

```
ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.2:testCompile (default-testCompile) on project easypost-java: Compilation failure
[ERROR] easypost-java/src/test/java/com/easypost/EasyPostTest.java:[154,60] diamond operator is not supported in -source 1.5
[ERROR] (use -source 7 or higher to enable diamond operator)
```

It is extremely simple to fix, but the question is whether someone have decided to do so on purpose or not...

jontsai:
Java SE 7 is now EOL'ed.

PR #21, opened 2016-01-20
Fix for a problem with source code version mismatch - upgrading to 1.7

uded:
As noted in #20


Justintime50:
This PR and #20 may not be needed anymore as the current version is `1.8`? @jontsai

jontsai:
Java SE 7 is now EOL'ed.

Issue #30, opened 2016-10-13
Compilation error


glapa:
[INFO] 11 errors 
[INFO] -------------------------------------------------------------
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 9.346 s
[INFO] Finished at: 2017-01-13T13:30:37+00:00
[INFO] Final Memory: 23M/282M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.5.1:testCompile (default-testCompile) on project easypost-api-client: Compilation failure: Compilation failure:
[ERROR] /home/jitpack/build/src/test/java/com/easypost/EasyPostTest.java:[453,34] cannot find symbol
[ERROR] symbol:   method containsString(java.lang.String)
[ERROR] location: class com.easypost.EasyPostTest
[ERROR] /home/jitpack/build/src/test/java/com/easypost/EasyPostTest.java:[454,34] cannot find symbol
[ERROR] symbol:   method containsString(java.lang.String)
[ERROR] location: class com.easypost.EasyPostTest
[ERROR] /home/jitpack/build/src/test/java/com/easypost/EasyPostTest.java:[455,34] cannot find symbol
[ERROR] symbol:   method containsString(java.lang.String)
[ERROR] location: class com.easypost.EasyPostTest
[ERROR] /home/jitpack/build/src/test/java/com/easypost/EasyPostTest.java:[533,5] cannot find symbol
[ERROR] symbol:   method assertNotEquals(java.lang.String,java.lang.String)
[ERROR] location: class com.easypost.EasyPostTest
[ERROR] /home/jitpack/build/src/test/java/com/easypost/EasyPostTest.java:[534,5] cannot find symbol
[ERROR] symbol:   method assertNotEquals(java.lang.String,java.lang.String)
[ERROR] location: class com.easypost.EasyPostTest
[ERROR] /home/jitpack/build/src/test/java/com/easypost/EasyPostTest.java:[684,74] cannot find symbol
[ERROR] symbol:   method containsString(java.lang.String)
[ERROR] location: class com.easypost.EasyPostTest
[ERROR] /home/jitpack/build/src/test/java/com/easypost/EasyPostTest.java:[689,75] cannot find symbol
[ERROR] symbol:   method containsString(java.lang.String)
[...]

Issue #15, opened 2015-05-06
Carrier Account Deleting needs to be added


rajsid1:
[Carrier Account Deleting](https://www.easypost.com/docs/api#carrier-accounts-delete)

rajsid1:
Currently, carrier can only be deleted after retrieving it. There should be a static method to delete a carrier by carrierId. It will save an API call.

Justintime50:
Closing due to inactivity. All of our client libraries follow the pattern of 1) retrieve record and 2) manipulate record, I don't believe we'll want to change this functionality at this time.

Please view our up-to-date [documentation](https://www.easypost.com/docs/api/java#carrier-accounts-delete-codesample) on deleting carrier accounts for more information.

PR #34, opened 2017-06-09
Updated CarrierAccount delete by ID

algreen11:
No longer have to retrieve CarrierAccount before you can delete #15 

pom changed to 4.11 #30


algreen11:
The java documentation on easypost needs to change. 

It is currently :
import com.easypost.EasyPost;

EasyPost.apiKey = "<YOUR_PRODUCTION_API_KEY>";

CarrierAccount ca = CarrierAccount.retrieve("ca_...");
ca.delete();

but can now be updated to :
import com.easypost.EasyPost;

EasyPost.apiKey = "<YOUR_PRODUCTION_API_KEY>";

CarrierAccount.delete("ca_...");

pom.xml
```diff
@@ -49,7 +49,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.10</version>
+            <version>4.11</version>
```
victoryftw: Are we intentionally trying to increase the version of Junit? That sounds like a different kind of change than the rest of whats done here. I think you want to be the version on line 8 instead.

src/main/java/com/easypost/model/CarrierAccount.java
```diff
@@ -57,11 +57,16 @@ public CarrierAccount update(Map<String, Object> params, String apiKey) throws E
...
+			if(id.substring(0,2).equals("ca")){
+				return delete(id,null); 
+			}
+			else{
+				return null;
+			}
 		}
-		public void delete(String apiKey) throws EasyPostException {
-			request(RequestMethod.DELETE, instanceURL(CarrierAccount.class, this.getId()), null, CarrierAccount.class, apiKey);
+		public static CarrierAccount delete(String id, String apiKey) throws EasyPostException {
```
victoryftw: I believe we still want this to be a `void` method, not returning a `CarrierAccount`. Once the carrier account is deleted, what is there to return?

src/main/java/com/easypost/model/CarrierAccount.java
```diff
@@ -57,11 +57,16 @@ public CarrierAccount update(Map<String, Object> params, String apiKey) throws E
...
+			return request(RequestMethod.DELETE, instanceURL(CarrierAccount.class, id), null, CarrierAccount.class, apiKey);
```
victoryftw: The CarrierAccount::Delete endpoint in EasyPost returns empty JSON, so I don't actually think there's anything here to return.

pom.xml
algreen11: This fixes a compilation error (#30) when initially building the library (at least it did for me).

algreen11:
Thats a good idea, I think we should keep both. Will update it now.

PR #35, opened 2017-06-28
Updated delete method fixes and added support for standalone rating

algreen11:
Can now use standalone rating ex: Rating rt = Rating.create(ratingMap);


pom.xml
```diff
@@ -49,7 +49,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.10</version>
+            <version>4.11</version>
```
victoryftw: Do we want to update the JUnit version? Is this intentional? I normally like to avoid package changes if possible, but if we need to bump it that's fine

src/main/java/com/easypost/EasyPost.java
```diff
@@ -3,7 +3,7 @@
 public abstract class EasyPost
 {
   public static String API_BASE = "https://api.easypost.com/v2";
-  public static final String VERSION = "3.3.5";
+  public static final String VERSION = "3.3.6";
```
victoryftw: 3.4.0

src/main/java/com/easypost/model/CarrierAccount.java
```diff
@@ -58,11 +58,19 @@ public CarrierAccount update(Map<String, Object> params, String apiKey) throws E
 
 		// delete
 		public void delete() throws EasyPostException {
-			this.delete(null);
+			this.delete(null, null);
 		}
-		public void delete(String apiKey) throws EasyPostException {
+		public void delete(String apiKey, String placeholder) throws EasyPostException {
```
victoryftw: Is this placeholder string necessary for something? Same with below, I feel like I'm missing the point

src/main/java/com/easypost/model/Rating.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.easypost.exception.EasyPostException;
+import com.easypost.net.EasyPostResource;
+
+public class Rating extends EasyPostResource {
+	List<?> ratings;
+	List<ShipmentMessage> messages;
+
+	public List<?>  getRating() {return ratings;}
+	public void setRatingList() {this.ratings = ratings;}
```
victoryftw: I think this should just be called "setRatings" if the field is called "ratings"

src/main/java/com/easypost/model/Rating.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.HashMap;
+
+import com.easypost.exception.EasyPostException;
+import com.easypost.net.EasyPostResource;
+
+public class Rating extends EasyPostResource {
+	List<?> ratings;
+	List<ShipmentMessage> messages;
+
+	public List<?>  getRating() {return ratings;}
```
victoryftw: `getRatings`? (plural)

pom.xml
algreen11: Yes it is intentional, one of the tests was failing on the initial build of the library, this fixed it for me

src/main/java/com/easypost/model/CarrierAccount.java
algreen11: This is tricky bc of the two methods for delete() and delete(id (and there subsequent delete functions which are called). I could just rename the added function to delete_id(id) to get rid of this String placeholder, but I thought that it would be more convenient to be able to call delete() and delete(id).

algreen11:
Let me know what you want me to do about the String placeholder issue and I'll edit it.

src/main/java/com/easypost/model/CarrierAccount.java
victoryftw: hmm yeah I see this issue now; since we want to have both static and instance methods here there's no good way to do this. Java obviously doesn't allow us to have static and instance methods with the same signature, and I don't really like this placeholder solution (since it would require users to pass in empty strings everywhere). I personally think the thing to do would be to make the static methods called something different:  either `destroy` or `deleteById` seem workable to me.

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -221,11 +221,16 @@ private static String urlEncodePair(String k, String v) throws UnsupportedEncodi
...
+	private static javax.net.ssl.HttpsURLConnection createPostConnection(String url, String query, String apiKey, Boolean isRating) throws IOException {
 		javax.net.ssl.HttpsURLConnection conn = createEasyPostConnection(url, apiKey);
 		conn.setDoOutput(true);
 		conn.setRequestMethod("POST");
-		conn.setRequestProperty("Content-Type", String.format("application/x-www-form-urlencoded;charset=%s", CHARSET));
+		if (!isRating) {
+			conn.setRequestProperty("Content-Type", String.format("application/x-www-form-urlencoded;charset=%s", CHARSET));
+		}
+		else{
+			conn.setRequestProperty("Content-Type", String.format("application/json"));
```
victoryftw: I'm not a massive fan of this `isRating` change, but if we need to tinker with the internals like this then I guess there's not much to do about. This conditional is the primary purpose of this variable, right?

algreen11: Yea, its because the request has to be made in json for Rating calls. If you have any thoughts on how to make this better definitely let me know.

Justintime50:
Standalone rating is no longer supported. Thank you for your efforts on this PR!

PR #44, opened 2018-03-14
add verification details and support shipment message as an object

ShobhaRaniD:
Add address verification details and additional error fields 
FedExSameDayCity returns rate error message as an object. Modifying "message" field to Object will support both string and object -


ShobhaRaniD:
Format files to conform to Google java style using this tool - https://github.com/google/google-java-format.

src/main/java/com/easypost/model/ShipmentMessage.java
```diff
@@ -3,26 +3,29 @@
 public class ShipmentMessage {
   String carrier;
   String type;
-  String message;
+  Object message;
```
mwaldt: Changing this to an object would not be backwards compatible and thus this would not be a minor version bump. This should at least be a mid version bump but I am more in favor of a major version bump.

ShobhaRaniD: I think a mid version bump would be applicable here, since historically we've had major version bumps if there were any changes in the dependent libraries.

ShobhaRaniD:
Bumped to 4.0.0, since I ended up adding two more fixes - handle "null" for rate currency, and send "start_date" and "end_date" in the top-level JSON of Report request params.

ShobhaRaniD:
If there are no more changes required, I'll proceed with merging the branch?

ShobhaRaniD:
👍

PR #56, opened 2020-10-19
send PUT requests via application/x-www-form-urlencoded body instead of query parameters

Roguelazer:
Literally just copy-pasta of the POST method.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
             <name>EasyPost Engineering</name>
             <email>oss@easypost.com</email>
             <organization>EasyPost</organization>
-            <organizationUrl>http://www.easypost.com</organizationUrl>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -108,6 +102,9 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-javadoc-plugin</artifactId>
+                <configuration>
+                    <source>8</source>
```
Justintime50, line 106: Hallelujah! This has been a problem for awhile.

Justintime50 (approved):
I know my review carries no weight but this LGTM. 

The configuration block is great, always had to manually add it - for at least the last year.

skrobot (approved):
Looks good. Tested the library with a simple Java application

PR #60, opened 2020-10-30
Removing containers and items objects, updating README

Justintime50:
It's been a hot minute since we supported `containers` or `items` from my understanding. Removing those here and updating README to have info on running tests instead of `coming soon`.

Wanted to get this in before a new major version was cut since these would require another bump anyway.


jontsai (approved):
LGTM

Sarahbear1926 (approved):

jontsai (approved):

PR #61, opened 2021-01-08
add github actions workflow; fix broken test

Roguelazer:


Roguelazer:
So, travis doesn't actually _run_ the tests, because it seems that they don't work. Grumble grumble.

Justintime50 (approved):
Sad day for tests

PR #63, opened 2021-02-03
use the new rerate api for rerating

Roguelazer:
Starting soon in the future, getRates() will always be read-only (as you would expect from a GET), and re-rating will be triggered through the newRates() function. This adjusts the newRates function. No change has occurred to getRates() yet.


Justintime50 (approved):
Looks fine, wish we had a test suite.

jontsai (approved):
LGTM

Issue #62, opened 2021-01-19
Run tests in CI


Roguelazer:
Our CI currently runs with `-DskipTests=true` because some of the tests don't work, and many of those that _do_ work aren't safe to run concurrently (which is required to test multiple versions of Java).

We should look into whether a good response mocking option a la vcr exists for whatever Java HTTP libraries we're using.

Justintime50:
@nwithan8 once VCR is done, let's enable tests on CI and close this out.

PR #67, opened 2021-10-04
style: format all files for better readability

Justintime50:
This repo is incredibly difficult to navigate due to the various unused imports, varying indent and organization spacing, etc

* Styles each file to use 2 spaces for consistency (this is the preferred spacing convention if following Google's style guide)
* Breaks up single function lines into multi-line for better readability
* Removes unused imports in each file
* Removes hard coded API Keys from README and example files (tests still have this present for now, will remove those in an upcoming PR).
* Replaces a `Boolean` deprecation in favor of its new usage
* Removes unused `.travis.yml` file

I will eventually put up another PR to add a linter to the project along with fixing the remaining lint errors, but this is a good initial step as we start beefing up this client library and need to be able to edit it easily.

### DIFF
--- a/src/main/java/com/easypost/net/EasyPostResource.java
+++ b/src/main/java/com/easypost/net/EasyPostResource.java
@@ -7,7 +7,6 @@ import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Field;
-import java.lang.reflect.Type;
 import java.lang.reflect.Method;
 import java.net.MalformedURLException;
 import java.net.URL;
@@ -24,10 +23,8 @@ import java.util.Date;
 import com.google.gson.FieldNamingPolicy;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
-import com.google.gson.reflect.TypeToken;
 import com.easypost.EasyPost;
 import com.easypost.exception.EasyPostException;
-import com.easypost.model.BatchStatus;
 import com.easypost.model.Event;
 import com.easypost.model.EventDeserializer;
 import com.easypost.model.Fee;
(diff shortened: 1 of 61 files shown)


src/test/java/com/easypost/UserTest.java
```diff
@@ -1,64 +1,45 @@
...
+ * 
+ * @Test public void testEditUser() throws EasyPostException { User me =
+ * User.retrieveMe();
+ * 
+ * Map<String, Object> params = new HashMap<String, Object>();
+ * params.put("name", "New Name2736"); me.update(params);
+ * 
+ * assertEquals("User was not updated successfully", "New Name2736",
+ * me.getName()); System.out.println(me.prettyPrint()); } }
+ */
```
fiendish, line 45: This formatting is not an improvement. (Why is this entire file commented out?)

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -38,573 +35,576 @@
...
+  }
+
+  // Batch
+  public List<Shipment> getShipments() {
+    return new ArrayList<Shipment>();
+  }
+
+  // public BatchStatus getStatus() {
+  // return new BatchStatus();
+  // }
```
fiendish, line 588: This formatting is not an improvement.

jontsai:
I like the 2 spaces for the XML files, but I'm used to 4 spaces for Java code (how I initially learned it). In fact, I'm generally accustomed to 4 spaces for every language.

Could you include a link to the Google Java Style guide? I'd like read it to better understand their rationale. As I understand, Google has a substantial amount of Java code (Android is completely Java, and even non-Android), so there's a chance that either my knowledge is way outdated, or that they are marching to the beat of their own drum.

Also, before making decisions on line-spacing, I think it would be meaningful to adopt [EditorConfig](https://editorconfig.org/) as a base standard, and first have a discussion on what settings should go inside of a `.editorconfig` file, before applying changes.

src/test/java/com/easypost/UserTest.java
Justintime50: The tests in this repo are VERY sad and need some love. Will happen in a follow up diff. We've commented out tests for now that we want to revisit. I completely agree that massive commented out blocks of code should just be removed.

Justintime50: For more context: https://github.com/EasyPost/easypost-java/issues/62

Justintime50: Like the Python CL, I'll need to blow away these changes once SmartRate lands anyway so I'll make sure to keep an eye out at that time. Switched this to draft until then.

Justintime50:
I've determined there is a LOT happening in this PR, we are going to handle everything in this PR but split them up into smaller ones so reviews are easier. I'll revisit your points @fiendish and @jontsai in those PRs.

PR #70, opened 2021-12-01
fix: Smartrate deserialization

nwithan8:
- adds new SmartrateCollection object to collect a list of Rate objects
- adds new SmartrateDeserializer to handle parsing smartrate-specific response JSON schema
- utilizes RateDeserializer to parse Rate objects, store in SmartrateCollection object
- getSmartrates() function utilizes SmartrateCollection internally, returns list of Rate objects (no change in return type)

I am not sure about testing procedures yet, so this is just a draft

### DIFF
--- a/src/main/java/com/easypost/model/Shipment.java
+++ b/src/main/java/com/easypost/model/Shipment.java
@@ -291,7 +291,6 @@ public class Shipment extends EasyPostResource {
 	}
 
 	// get smartrates
-
     public List<Rate> getSmartrates() throws EasyPostException {
         return this.getSmartrates(null, null);
     }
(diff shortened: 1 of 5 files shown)


src/main/java/com/easypost/model/Shipment.java
```diff
@@ -291,7 +291,6 @@ public Shipment newRates(Map<String, Object> params, String apiKey) throws EasyP
 	}
 
 	// get smartrates
```
Justintime50, line 293: Can we unindent this?

Justintime50 (changes requested):

Justintime50 (approved):
Code LGTM, tested with the team and it works properly.

This will merge into @jontsai's branch to complete it and is necessary to move forward with SmartRate.

Sarahbear1926 (approved):

nwithan8:
Re-added some things that got lost in the b88fe51 rollback.
Tests run correctly locally.

ajnassar (approved):

jchen293 (approved):
LGTM, tested it with the team and the code works as we expected.

PR #73, opened 2021-12-03
remove travis.yml file from the Java repo because we no longer use tr…

jchen293:
…avis as it was accidently leftover


Justintime50 (approved):
We don't use travis anymore, needs to go.

nwithan8:
Looks good

PR #74, opened 2021-12-08
Tax identifier

jchen293:
Add tax identifier for Java Client Library

### DIFF
--- a/src/main/java/com/easypost/model/Shipment.java
+++ b/src/main/java/com/easypost/model/Shipment.java
@@ -35,6 +35,7 @@ public class Shipment extends EasyPostResource {
 	String uspsZone;
 	Map<String, Object> options;
 	List<ShipmentMessage> messages;
+    List<TaxIdentifier> taxIdentifiers;
 
 	public String getId() {
 		return id;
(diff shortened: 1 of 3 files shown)


src/main/java/com/easypost/model/Shipment.java
```diff
@@ -35,6 +35,7 @@
 	String uspsZone;
 	Map<String, Object> options;
 	List<ShipmentMessage> messages;
+    List<TaxIdentifier> taxIdentifiers;
```
Justintime50, line 38: This is indented incorrectly, can we fix it? Also, it should be singular.
```suggestion
    List<TaxIdentifier> taxIdentifier;
```

src/main/java/com/easypost/model/Shipment.java
```diff
@@ -225,6 +226,13 @@ public void setForms(List<Form> forms) {
 		this.forms = forms;
 	}
 
+	public List<TaxIdentifier> getTaxIdentifier(){
```
Justintime50, line 229:
```suggestion
	public List<TaxIdentifier> getTaxIdentifiers(){
```

Justintime50:
This still wasn't rebased correctly, it contains the already removed `.travis.yml` file removal which was already done.

This will need to be rebase properly.

jchen293:
Created the taxIdentifier branch for PR.

PR #75, opened 2021-12-10
Migrate from urlencoded to JSON bodies for POST/PUT requests

nwithan8:
Java library will now send data as JSON in the request body for POST and PUT requests, rather than sending them as urlencoded data.
Still uses urlencoded data for GET and DELETE functions.

This allows for future support for things such as Tax Identifiers.

### DIFF
--- a/src/main/java/com/easypost/model/Tracker.java
+++ b/src/main/java/com/easypost/model/Tracker.java
@@ -10,159 +10,159 @@ import java.util.List;
 import java.util.Map;
 
 public class Tracker extends EasyPostResource {
-	public String id;
-	String mode;
-	String trackingCode;
-	String status;
-	String shipmentId;
-	String carrier;
-	List<TrackingDetail> trackingDetails;
-	float weight;
-	Date estDeliveryDate;
-	String signedBy;
-	CarrierDetail carrierDetail;
-	String publicUrl;
-	String statusDetail;
+    public String id;
+    String mode;
+    String trackingCode;
+    String status;
+    String shipmentId;
+    String carrier;
+    List<TrackingDetail> trackingDetails;
+    float weight;
+    Date estDeliveryDate;
+    String signedBy;
+    CarrierDetail carrierDetail;
+    String publicUrl;
+    String statusDetail;
 
-	public String getId() {
-		return id;
-	}
-	public void setId(String id) {
-		this.id = id;
-	}
+    public String getId() {
+        return id;
+    }
+    public void setId(String id) {
+        this.id = id;
+    }
 
-	public String getMode() {
-		return mode;
-	}
-	public void setMode(String mode) {
-		this.mode = mode;
-	}
+    public String getMode() {
+        return mode;
+    }
+    public void setMode(String mode) {
+        this.mode = mode;
+    }
 
-	public String getShipmentId() {
-		return shipmentId;
-	}
-	public void setShipmentId(String shipmentId) {
-		this.shipmentId = shipmentId;
-	}
+    public String getShipmentId() {
+        return shipmentId;
+    }
+    public void setShipmentId(String shipmentId) {
+        this.shipmentId = shipmentId;
+    }
 
-	public String getCarrier() {
-		return carrier;
-	}
-	public void setCarrier(String carrier) {
-		this.carrier = carrier;
-	}
+    public String getCarrier() {
+        return carrier;
+    }
+    public void setCarrier(String carrier) {
+        this.carrier = carrier;
+    }
 
-	public String getTrackingCode() {
-		return trackingCode;
-	}
-	public void setTrackingCode(String trackingCode) {
-		this.trackingCode = trackingCode;
-	}
+    public String getTrackingCode() {
+        return trackingCode;
+    }
+    public void setTrackingCode(String trackingCode) {
+        this.trackingCode = trackingCode;
+    }
 
-	public String getStatus() {
-		return status;
-	}
-	public void setStatus(String status) {
-		this.status = status;
-	}
+    public String getStatus() {
+        return status;
+    }
+    public void setStatus(String status) {
+        this.status = status;
+    }
 
-	public List<TrackingDetail> getTrackingDetails() {
-		return trackingDetails;
-	}
-	public void setTrackingDetails(List<TrackingDetail> trackingDetails) {
-		this.trackingDetails = trackingDetails;
-	}
+    public List<TrackingDetail> getTrackingDetails() {
+        return trackingDetails;
+    }
+    public void setTrackingDetails(List<TrackingDetail> trackingDetails) {
+        this.trackingDetails = trackingDetails;
+    }
 
-	public float getWeight() {
-		return weight;
-	}
-	public void setWeight(float weight) {
-		this.weight = weight;
-	}
+    public float getWeight() {
+        return weight;
+    }
+    public void setWeight(float weight) {
+        this.weight = weight;
+    }
 
-	public Date getEstDeliveryDate() {
-		return estDeliveryDate;
-	}
-	public void setEstDeliveryDate(Date estDeliveryDate) {
-		this.estDeliveryDate = estDeliveryDate;
-	}
+    public Date getEstDeliveryDate() {
+        return estDeliveryDate;
+    }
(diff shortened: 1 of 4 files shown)


Justintime50 (commented):
Overall this looks pretty good. We can sync about the couple questions I left. I'd like to also run some tests of my own before approving. Won't block yet.

src/main/java/com/easypost/model/Tracker.java
```diff
@@ -148,7 +149,16 @@ public static boolean createList(Map<String, Object> params) throws EasyPostExce
...
+		Map<String, Object> newParams = new HashMap<String, Object>();
+		Map<String, Object> trackers = new HashMap<String, Object>();
+		for (Object tracker : (ArrayList)params.get("trackers")) {
+			trackers.put(String.valueOf(count), tracker);
+			count++;
+		}
+		newParams.put("trackers", trackers);
+
+		request(RequestMethod.POST, createListUrl, newParams, Object.class, apiKey);
 		return true;
```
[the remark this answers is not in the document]
Justintime50: Disregard my concern, I'm now noticing this is specific to the `createList` method already, I initially thought this was in the "body builder" and didn't want it there. 

I do want to explore if we want to fix this on the API side or if we'll need to fix this on all the client libraries instead. Let's review everything aside from this block here and we'll circle back shortly.

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -389,7 +395,7 @@ private static EasyPostResponse makeURLConnectionRequest(EasyPostResource.Reques
 
 	protected static <T> T request(EasyPostResource.RequestMethod method, String url, Map<String, Object> params, Class<T> clazz, String apiKey, boolean apiKeyRequired) throws EasyPostException {
 		String originalDNSCacheTTL = null;
-		Boolean allowedToSetTTL = true;
+		boolean allowedToSetTTL = true;
```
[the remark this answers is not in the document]
Justintime50: Let's leave this for now, I know there are a bunch of optimizations and deprecated methods we can update, let's visit them all together in a separate PR.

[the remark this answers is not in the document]
nwithan8: Rolled back in 85c0741

src/main/java/com/easypost/model/Tracker.java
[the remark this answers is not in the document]
Justintime50: After talking internally, we may also allow this endpoint to accept a list of JSON objects, but for now we'll move forward with the hash of objects as it currently accepts this.

Justintime50 (approved):
Code lgtm, end to end tests were done on shipments, addresses, parcels, tax identifiers, customs items, trackers. Various data types were tried such as floats, bools, null, strings, lists, etc according to @nwithan8.

This will require a major version bump as it's a breaking change.

Justintime50 (approved):

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -429,25 +435,38 @@ private static EasyPostResponse makeURLConnectionRequest(EasyPostResource.Reques
...
+					try {
+						query = createQuery(params);
+					} catch (UnsupportedEncodingException e) {
+						throw new EasyPostException(
+								String.format("Unable to encode parameters to %s. Please email %s for assistance.", CHARSET, EasyPostResource.EASYPOST_SUPPORT_EMAIL), e);
+					}
+					break;
+				case POST:
+				case PUT:
+					body = createBody(params);
```
jchen293, line 453: Should we add a try-catch block here?

src/main/java/com/easypost/model/Tracker.java
```diff
@@ -148,7 +149,16 @@ public static boolean createList(Map<String, Object> params) throws EasyPostExce
 	public static boolean createList(Map<String, Object> params, String apiKey) throws EasyPostException {
 		String createListUrl = String.format("%s/create_list", classURL(Tracker.class));
 
-		request(RequestMethod.POST, createListUrl, params, Object.class, apiKey);
+		int count = 0;
+		Map<String, Object> newParams = new HashMap<String, Object>();
+		Map<String, Object> trackers = new HashMap<String, Object>();
+		for (Object tracker : (ArrayList)params.get("trackers")) {
```
jchen293, line 155: Add a space after casting (ArrayList).
If a user inputs value that is not a list for the tracker, the casting would fail. Should we add a try-catch block to handle this?

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -281,6 +284,11 @@ private static String createQuery(Map<String, Object> params) throws Unsupported
 		return queryStringBuffer.toString();
 	}
 
+	private static JsonObject createJsonObjectFromMap(Map<String, Object> map) {
```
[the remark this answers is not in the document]
nwithan8: It could be, createJsonObjectFromMap() is a generic method, while createBody() is the specific implementation for HTTP requests.
But we could probably just remove the generi createJsonObjectFromMap() because 1) it's probably not right to have a generic method in this file and 2) it's not being used by anything else, so it's actually just adding more lines rather than saving any lines at the moment.

src/main/java/com/easypost/net/EasyPostResource.java
nwithan8: We could. The createBody() handles null values upstream, but for other errors that might arise, good idea.

src/main/java/com/easypost/model/Tracker.java
nwithan8: Sounds good.

nwithan8: Will address formatting in a separate branch.

src/main/java/com/easypost/model/Tracker.java
```diff
@@ -146,10 +147,22 @@ public static boolean createList(Map<String, Object> params) throws EasyPostExce
...
-
-		request(RequestMethod.POST, createListUrl, params, Object.class, apiKey);
-		return true;
+		try {
+			String createListUrl = String.format("%s/create_list", classURL(Tracker.class));
+
+			int count = 0;
+			Map<String, Object> newParams = new HashMap<String, Object>();
+			Map<String, Object> trackers = new HashMap<String, Object>();
+			for (Object tracker : (ArrayList)params.get("trackers")) {
```
jchen293, line 156: I would add a space after casting ArrayList

src/main/java/com/easypost/model/Tracker.java
```diff
@@ -146,10 +147,22 @@ public static boolean createList(Map<String, Object> params) throws EasyPostExce
...
+				trackers.put(String.valueOf(count), tracker);
+				count++;
+			}
+			newParams.put("trackers", trackers);
+
+			request(RequestMethod.POST, createListUrl, newParams, Object.class, apiKey);
+			return true;
+		} catch (Exception e) {
+			return false;
```
jchen293, line 165: Can we throw an EasyPostException before the **return false** so the user can get a message?

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -251,26 +260,21 @@ private static String urlEncodePair(String k, String v) throws UnsupportedEncodi
...
 
+	private static JsonObject createBody(Map<String, Object> params) {
+		Gson gson = new Gson();
+		return gson.toJsonTree(params).getAsJsonObject();
+	}
+
 	private static String createQuery(Map<String, Object> params) throws UnsupportedEncodingException {
 		Map<String, String> flatParams = flattenParams(params);
-		StringBuffer queryStringBuffer = new StringBuffer();
+		StringBuilder queryStringBuffer = new StringBuilder();
```
jchen293, line 277: Rename **queryStringBuffer** to **queryStringBuilder**

jchen293 (approved):
LGTM, according to @nwithan8 he has done a lot of testing on various data input types with the tool he built. I also did some testing locally and haven't found any errors.

Justintime50 (approved):

jontsai (approved):

PR #77, opened 2021-12-10
Fixed hard code for support mail 

jchen293:
Added a variable for support mail and replaced it with hard-coded emails with the constant inline with the strings using it.
Removed print statement.

### DIFF
--- a/src/main/java/com/easypost/model/EventDeserializer.java
+++ b/src/main/java/com/easypost/model/EventDeserializer.java
@@ -70,8 +70,7 @@ public class EventDeserializer implements JsonDeserializer<Event> {
 		} else if (element.isJsonArray()) {
 			return deserializeJsonArray(element.getAsJsonArray());
 		} else {
-			System.err.println("Unknown JSON element type for element " + element + ". " +
-					"Please email us at " + EasyPostResource.EASYPOST_SUPPORT_EMAIL + ".");
+			System.err.println(String.format("Unknown JSON element type for element %s. Please email us at %s.", element.toString(), EasyPostResource.EASYPOST_SUPPORT_EMAIL));
 			return null;
 		}
 	}
--- a/src/main/java/com/easypost/model/EventDeserializer.java
+++ b/src/main/java/com/easypost/model/EventDeserializer.java
@@ -71,7 +71,7 @@ public class EventDeserializer implements JsonDeserializer<Event> {
 			return deserializeJsonArray(element.getAsJsonArray());
 		} else {
 			System.err.println("Unknown JSON element type for element " + element + ". " +
-					"Please email us at support@easypost.com.");
+					"Please email us at " + EasyPostResource.EASYPOST_SUPPORT_EMAIL + ".");
 			return null;
 		}
 	}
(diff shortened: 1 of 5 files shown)


src/test/java/com/easypost/EasyPostTest.java
```diff
@@ -851,15 +851,6 @@ public void testCustomsInfo() throws EasyPostException {
...
     shipment = shipment.buy(shipment.lowestRate(buyCarriers));
-
-    ///System.out.println("Shipment.getForms(): >>"+shipment.getForms()+"<<");
-    /// java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
-    /// Form does not exists. Invalid test
-    /*
-    assertTrue(
-        "Customs Form was not submitted electronically",
-        shipment.getForms().get(0).getSubmittedElectronically());
-        */
```
Justintime50, line 862: Let's leave code alone that isn't part of the ask for the PR.

src/main/java/com/easypost/net/EasyPostResponse.java
```diff
@@ -8,8 +8,6 @@
 	public EasyPostResponse(int responseCode, String responseBody) {
 		this.responseCode = responseCode;
 		this.responseBody = responseBody;
-
-    // System.out.println(this.responseBody);
```
Justintime50, line 12: This is already being removed from the JSON PR @nwithan8 is doing, you can leave it alone and it'll be take care of separately.

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -583,17 +584,12 @@ public String getId() {
 	public String getMode() {
 		return "";
 	}
-	// Batch
 	public List<Shipment> getShipments() {
 		return new ArrayList<Shipment>();
 	}
-	// public BatchStatus getStatus() {
-	// 	return new BatchStatus();
-	// }
```
Justintime50, line 592: Let's leave code alone that isn't part of the ask for the PR.

Justintime50 (changes requested):

nwithan8 (approved):
LGTM

Justintime50 (commented):
Ah yes... string interpolation - one of many things Java doesn't have. Would we want to consider using `String.format()` instead to clean it up so we don't have to break up the strings?

jchen293:
Used String.format() to clean up the print statement.

src/main/java/com/easypost/model/EventDeserializer.java
```diff
@@ -71,7 +71,7 @@ private Object deserializeJsonElement(JsonElement element) {
 			return deserializeJsonArray(element.getAsJsonArray());
 		} else {
 			System.err.println("Unknown JSON element type for element " + element + ". " +
-					"Please email us at support@easypost.com.");
+					"Please email us at " + EasyPostResource.EASYPOST_SUPPORT_EMAIL + ".");
```
Justintime50, line 74: Looks like there was one missed to switch to `String.format()`

Justintime50 (changes requested):
The rest looks good, one was missed.

Justintime50 (approved):

nwithan8 (approved):

PR #79, opened 2021-12-17
Provide enums for values with set of options

nwithan8:
Provide enums for fields that have only a specific set of choices for values (i.e. "pickup status" can only be one of four or five different strings).
This makes it easier to set values, since it requires the user to use only "approved" values, rather than any potential string.
This makes it easier to get and use values, since it will ensure that the user received one of an "approved" or expected value.  This naturally makes it easier to compare/check results by avoiding string-to-string comparisons, which are prone to spelling errors.
This also helps expose if an error does occur (i.e. the API returned unexpected data and subsequently throws a deserialization error).
Enums can also be marked as "deprecated" when old values are being phased out.
Conversely, anything not explicitly included as an enum choice will fail (programmatically produces null) if the API returns the data. This would require us to update the Java library often to include the new options, i.e. whenever a new carrier is added.

### DIFF
--- a/src/main/java/com/easypost/app/BatchManifestExample.java
+++ b/src/main/java/com/easypost/app/BatchManifestExample.java
@@ -4,6 +4,7 @@ import com.easypost.EasyPost;
 import com.easypost.exception.EasyPostException;
 import com.easypost.model.Batch;
 import com.easypost.model.Shipment;
+import com.easypost.model.enums.BatchState;
 
 import java.util.ArrayList;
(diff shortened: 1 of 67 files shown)


nwithan8:
This is just preliminary.

TODO:
- Verify the names and available choices for each enum
- Implement in all domain classes as needed

Justintime50:
Closing as this is wildly out of date and won't be included in the near future. Can always reference this when we circle back.

PR #81, opened 2021-12-29
Fix incorrect test value

nwithan8:
Fixes an incorrect expected value that will cause the test to fail. Must have gotten lost in a rebase.

### DIFF
--- a/src/test/java/com/easypost/EasyPostTest.java
+++ b/src/test/java/com/easypost/EasyPostTest.java
@@ -201,7 +201,7 @@ public class EasyPostTest {
     assertEquals("LabelFee", fee1.getType());
 
     Fee fee2 = fees.get(1);
-    assertEquals(4.83, fee2.getAmount(), 0.001);
+    assertEquals(4.53, fee2.getAmount(), 0.001);
     assertEquals(true, fee2.getCharged());
     assertEquals(false, fee2.getRefunded());
     assertEquals("PostageFee", fee2.getType());


Justintime50:
I'm confused why this is necessary unless this fee changed in the last couple hours. We just merged a commit without this fix and the tests passed fine earlier today.

jontsai (approved):

nwithan8:
> I'm confused why this is necessary unless this fee changed in the last couple hours. We just merged a commit without this fix and the tests passed fine earlier today.

¯\_(ツ)_/¯

All I know is, tests are failing with "4.83" on the master branch, and not failing with "4.53"
image

jchen293 (approved):
Code LGTM, we might want to confirm if the fee actually got increased before the merge.

fiendish:
Screen Shot 2021-12-30 at 4 03 30 PM

fiendish (approved):
This is ok, but checking for specific values against a server response where values change over time is not ideal.

nwithan8:
> This is ok, but checking for specific values against a server response where values change over time is clearly not ideal.

VCR is on our radar.

fiendish:
I'm also not a fan of VCR because it allows the client and server to desynchronize when someone fails to update the recordings. I'd rather test against a real server but not specific values.

Justintime50 (approved):
I guess, still confused why this broke in-between two PRs an hour apart, but this is a trivial change we can adjust if needed.

fiendish:
It will break any time service rates change.

PR #82, opened 2022-01-12
Add linter tools for enforcing Java style

nwithan8:
This branch adds two new tools for detecting violations of our newly-established EasyPost Java style guide (via CheckStyle):
- a new step to our `CI` GitHub Action to execute CheckStyle checks on any non-`master` push. 
- included Apache Maven CheckStyle plugin in `pom.xml`. Developers can check for violations by running `mvn verify`


nwithan8:
NOTE: The CheckStyle action is running on this PR, but since this code is based on the `master` branch, there is no `easypost_java_style.xml` file to use for the checks. This will be resolved when the `style` branch is merged into `master` and this branch is rebased.

dev_tools.sh
```diff
@@ -0,0 +1,42 @@
...
+  # First, check if CheckStyle JAR exists, and download if not
+  if [ ! -f $DEVTOOLS_FOLDER/checkstyle.jar ]
+  then
+    echo "Downloading missing CheckStyle JAR file..."
+    load
+  fi
+
+  # Download EasyPost CheckStyle files from GitHub (every time in case they've changed)
+  echo "Downloading EasyPost CheckStyle config files..."
+  wget -q --show-progress -O $DEVTOOLS_FOLDER/$STYLE_FILE https://raw.githubusercontent.com/EasyPost/easypost-java/style/easypost_java_style.xml
```
nwithan8, line 33: These two URLs currently point to the files on our `style` branch, because they do not exist on the `master` branch yet. We will update these when the `style` branch has been merged into `master`.

.github/workflows/checkstyle.yml
```diff
@@ -0,0 +1,20 @@
...
+  checkstyle_job:
+    runs-on: ubuntu-latest
+    name: Run CheckStyle checks
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Check for EasyPost style compliance
+        uses: nikitasavinov/checkstyle-action@master
+        with:
+          # github_token: ${{ secrets.GITHUB_TOKEN }}
```
Justintime50, line 18: Let's determine if this is needed or not, uncomment if it is, remove completely if not.

.github/workflows/checkstyle.yml
```diff
@@ -0,0 +1,20 @@
...
+
+on:
+  push:
+    branches: [ master ]
+  pull_request: ~
+
+jobs:
+  checkstyle_job:
+    runs-on: ubuntu-latest
+    name: Run CheckStyle checks
```
Justintime50, line 11: I'd remove this `name` key here completely and only use the ones below

.github/workflows/checkstyle.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Check for EasyPost style compliance
```
Justintime50, line 15: Nit: Could change to
```suggestion
      - name: Run Checkstyle
```

.gitignore
```diff
@@ -20,3 +20,4 @@ settings.xml
 
 
 src/scratch
+dev_tools/
```
Justintime50, line 23: After looking through, I realize this was for the jar files that get downloaded?

If so, I don't think we should include those scripts at this time and thus we won't need to ignore this dir.

dev_tools.sh
Justintime50: In theory, this shouldn't need to be downloaded though since it's included in the repo right?

pom.xml
```diff
@@ -162,6 +162,26 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
```
Justintime50, line 167: Is this addition to define Checkstyle as a dependency? If so, shouldn't it be made so with the `dev` scope?

dev_tools.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
Justintime50, line 1: Thanks for taking the initiative with the shell and batch scrips! After thinking on it, I feel more comfortable if we simply did this work with the Maven/Gradle adoption here soon so that we can just use the platform's specific programs to manage dependencies instead of through one-off scrips.

What do you think? We able to just allow users to manually download Checkstyle for now and we can revisit this when we adopt Maven/Gradle?

.github/workflows/checkstyle.yml
```diff
@@ -0,0 +1,20 @@
+name: CheckStyle
```
Justintime50, line 1: All of the content of this workflow could be placed as an additional step in the already existent `CI.yml` workflow. We don't have separate workflows for our other projects and include `test`, `lint`, `build` all in a single workflow. Let's do that here too.

Justintime50 (dismissed):
Let's remove the scripts for now and handle those in a separate PR and instead use Maven/Gradle to do dependency management. Other than that I had a question and a couple of nits on the action itself.

Justintime50:
This can also be rebased now that the style PR has been merged.

dev_tools.sh
nwithan8: That's true. I think my logic was, "let's always use the style guide from the master branch in case it's different than the one on this branch"

pom.xml
nwithan8: This makes CheckStyle a plugin for development (different from a dependency)

dev_tools.sh
nwithan8: I have already added the Maven plugin, so yes, we can remove these scripts.

.github/workflows/checkstyle.yml
nwithan8: Fixed in 5510fba

.github/workflows/checkstyle.yml
nwithan8: Fixed in 5510fba

.github/workflows/ci.yml
```diff
@@ -19,4 +19,13 @@ jobs:
           java-version: ${{ matrix.javaversion }}
       - name: Build with Maven
         run: mvn --batch-mode install -Dgpg.skip=true -DskipTests=true
+  checkstyle_job:
```
Justintime50, line 22: Nit: This should just be called `lint`

pom.xml
Justintime50: Gotcha. We appear to have a bunch of plugins already, this is probably fine.

.github/workflows/ci.yml
nwithan8: Fixed in a4d8c8a

Justintime50 (dismissed):
Thanks for this! Now this repo will stay pretty.

jontsai (dismissed):

Justintime50:
@nwithan8 can you actually pin this to [0.4.0](https://github.com/nikitasavinov/checkstyle-action/releases/tag/0.4.0) per external conversations?

jchen293 (approved):
lgtm!

Justintime50 (approved):

jchen293 (approved):
lgtm

PR #83, opened 2022-01-14
add Java version to the User-Agent in requests

jchen293:
Pass Java version to the User-Agent in requests


nwithan8 (approved):
LGTM

Sounds like this is guaranteed to work in every version of Java that we support (8+): https://stackoverflow.com/a/2591122/13343799

fiendish:
> Sounds like this is guaranteed to work in every version of Java that we support (8+):

Our README says "Java 1.5 and later."

fiendish (approved):

Justintime50 (approved):
@jchen293 did you test this? If so, can you add a screenshot or some other verifiable item?

PR #86, opened 2022-01-17
chore: bump plugins, remove bad Java README version

Justintime50:
Bumps the plugins we use during development, has no impact on the library itself. Also removes the now incorrect `Java 1.5` requirement from the README


jchen293 (approved):
LGTM

PR #88, opened 2022-01-24
add support to retrieve available Carrier Types

jchen293:
This PR adds support for users to retrieve available Carrier Types.

### DIFF
--- a/src/main/java/com/easypost/model/CarrierType.java
+++ b/src/main/java/com/easypost/model/CarrierType.java
@@ -20,82 +20,93 @@ public final class CarrierType extends EasyPostResource {
     private Map<String, Object> fields;
 
     /**
-     * @return type of a carrier
+     * Get the type of the carrier.
+     *
+     * @return type of a carrier.
      */
     public String getType() {
         return type;
     }
 
     /**
+     * Set the type of carrier.
      *
-     * @param type type of a carrier
+     * @param type type of a carrier.
      */
     public void setType(final String type) {
         this.type = type;
     }
 
     /**
+     * Get the readable of the carrier.
      *
-     * @return return readable from a carrier
+     * @return return readable from a carrier.
      */
     public String getReadable() {
         return readable;
     }
 
     /**
+     * Set the readable of carrier.
      *
-     * @param readable readable for a carrier
+     * @param readable readable for a carrier.
      */
     public void setReadable(final String readable) {
         this.readable = readable;
     }
 
     /**
+     * Get the logo of the carrier.
      *
-     * @return logo of a carrier
+     * @return logo of a carrier.
      */
     public String getLogo() {
         return logo;
     }
 
     /**
+     * Set the logo of carrier.
      *
-     * @param logo logo of a carrier
+     * @param logo logo of a carrier.
      */
     public void setLogo(final String logo) {
         this.logo = logo;
     }
 
     /**
+     * Get the fields of the carrier.
      *
-     * @return fields from a carrier
+     * @return fields from a carrier.
      */
     public Map<String, Object> getFields() {
         return fields;
     }
 
     /**
+     * Set the fields of carrier.
      *
-     * @param fields fields for a carrier
+     * @param fields fields for a carrier.
      */
     public void setFields(final Map<String, Object> fields) {
         this.fields = fields;
     }
 
     /**
+     * Retrieve a list of available carrier for the account.
      *
-     * @return list of carrier type that is available to the account
-     * @throws EasyPostException
+     * @return list of carrier type that is available to the account.
+     * @throws EasyPostException when the request fails.
      */
     public static List<CarrierType> all() throws EasyPostException {
         return all(null);
     }
 
     /**
+     * Retrieve a list of available carrier for the given account API key.
      *
-     * @param apikey the API key from the user input
-     * @return list of carrier type that is available to the account of the given API key
-     * @throws EasyPostException
+     * @param apikey the API key from the user input.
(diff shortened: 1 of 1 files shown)


Justintime50 (changes requested):
Please add a description to each function's docstring (see code in other classes for an example).

jontsai (approved):

src/main/java/com/easypost/model/CarrierType.java
```diff
@@ -0,0 +1,115 @@
...
+     * Set the type of carrier.
+     *
+     * @param type type of a carrier.
+     */
+    public void setType(final String type) {
+        this.type = type;
+    }
+
+    /**
+     * Get the readable of the carrier.
```
Justintime50, line 41:
```suggestion
     * Get the readable name of the carrier.
```

src/main/java/com/easypost/model/CarrierType.java
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @return return readable from a carrier.
```
Justintime50, line 43:
```suggestion
     * @return return readable name of the carrier.
```

src/main/java/com/easypost/model/CarrierType.java
```diff
@@ -0,0 +1,115 @@
...
+     */
+    public String getReadable() {
+        return readable;
+    }
+
+    /**
+     * Set the readable of carrier.
```
Justintime50, line 50:
```suggestion
     * Set the readable name of the carrier.
```

src/main/java/com/easypost/model/CarrierType.java
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @param readable readable for a carrier.
```
Justintime50, line 52:
```suggestion
     * @param readable readable name of the carrier.
```

src/main/java/com/easypost/model/CarrierType.java
```diff
@@ -0,0 +1,115 @@
...
+     * Set the fields of carrier.
+     *
+     * @param fields fields for a carrier.
+     */
+    public void setFields(final Map<String, Object> fields) {
+        this.fields = fields;
+    }
+
+    /**
+     * Retrieve a list of available carrier for the account.
```
Justintime50, line 95:
```suggestion
     * Retrieve a list of available carriers for the account.
```

src/main/java/com/easypost/model/CarrierType.java
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @return list of carrier type that is available to the account.
```
Justintime50, line 97:
```suggestion
     * @return list of carrier types that are available to the account.
```

src/main/java/com/easypost/model/CarrierType.java
```diff
@@ -0,0 +1,115 @@
...
+     * @throws EasyPostException when the request fails.
+     */
+    public static List<CarrierType> all() throws EasyPostException {
+        return all(null);
+    }
+
+    /**
+     * Retrieve a list of available carrier for the given account API key.
```
Justintime50, line 105:
```suggestion
     * Retrieve a list of available carriers for the given account.
```

src/main/java/com/easypost/model/CarrierType.java
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @param apikey the API key from the user input.
+     * @return list of carrier type that is available to the account of the given API key.
```
Justintime50, line 108:
```suggestion
     * @return list of carrier types that are available for the given account.
```

Justintime50 (changes requested):
Code looks fine, you'll want to try and find a consistent lexicon of phrases and use throughout the whole PR. I see a lot of switching between `of`, `for` and `a`, `the` when referencing the carrier accounts. Similarly there are some inconsistencies with pluralism that should be corrected.

We want to try being as consistent and clean as possible 🙂

nwithan8:
> Code looks fine, you'll want to try and find a consistent lexicon of phrases and use throughout the whole PR. I see a lot of switching between `of`, `for` and `a`, `the` when referencing the carrier accounts. Similarly there are some inconsistencies with pluralism that should be corrected.
> 
> We want to try being as consistent and clean as possible 🙂

Regarding this, if it's a static function, it would be getting `a/an` object or something relative to `a/an` object.
If it's a non-static function, relying on an instance of an object, then it's getting `the` object or something relative to `the` object.

nwithan8 (approved):

src/main/java/com/easypost/model/CarrierType.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Retrieve a list of available carriers for the given account.
+     *
+     * @param apikey the API key from the user input.
+     * @return list of carrier types that are available for the given account.
+     * @throws EasyPostException when the request fails.
+     */
+    public static List<CarrierType> all(final String apikey) throws EasyPostException {
+        CarrierType[] ret = request(RequestMethod.GET, classURL(CarrierType.class), null, CarrierType[].class, apikey);
```
Justintime50, line 112: Nit: let's call `ret` what it actually is which is `response`. Won't block on it.

Justintime50 (dismissed):

src/main/java/com/easypost/model/CarrierType.java
jchen293: resolved.

Justintime50 (approved):

nwithan8 (approved):

Issue #78, opened 2021-12-14
Migration to eventually fix spelling error?


nwithan8:
https://github.com/EasyPost/easypost-java/blob/cdd2d48333254a3a10200dace92b1fa6c0053d11/src/main/java/com/easypost/model/Tracker.java#L99

jchen293:
It calls the method in the EasyPostResource file, but I think we can have a better naming for the method to avoid confusion.

Justintime50:
Dang, should have included this in our last major version bump. Too bad. I'll file an internal task to accompany this so we can try to plan to include it whenever our next major version bump comes around.

PR #90, opened 2022-01-25
Migration to eventually fix spelling error

nwithan8:
Addresses #78 

- Remove misspelled `get/setUpdateAt()`, defer to internal `get/setUpdatedAt()`
- Fixed test


Justintime50:
Converting to a draft. We won't be moving on this in the near future since it's a breaking change and we just did a major version bump. It can hang around until we get to that point.

Justintime50:
Closing for now, we can revisit whenever we want; however, this is falling out of sync with the repo and isn't going to be fixed in the near future (waiting for the next major release).

PR #92, opened 2022-01-28
Linting (round 2)

nwithan8:
During our last round of linting, we were focused on variable names, `final` markers, CheckStyle enforcement and the like. But we didn't check for file layout (meaning in what order items in a file are organized); for example:
- static methods come before non-static methods
- final variables come before non-final variables
- constructors come before methods
- public methods come before private methods

Sample of rules:
image

Luckily, an IDE (at least, JetBrains IntelliJ) can automatically rearrange code to conform to such layout rules. This PR:
- rearranges code in all files to meet layout rules
- adds .editorconfig so these rules can be easily imported and applied via any developer's IDE

These rules for now are the default ones via JetBrains IntelliJ (likely based off a common community-agreed-upon layout), but can be altered if desired.

**No code was added or removed** during this process, simply reordered.


src/main/java/com/easypost/model/Shipment.java
```diff
@@ -552,7 +625,7 @@ public void setTaxIdentifiers(final List<TaxIdentifier> taxIdentifiers) {
 
     /**
      * Get the carrier accounts of this shipment.
-     *
+     * <p>
```
Justintime50, line 628: This doesn't seem necessary? Accidentally added?

src/main/java/com/easypost/model/Tracker.java
```diff
@@ -157,6 +265,8 @@ public void setTrackingDetails(final List<TrackingDetail> trackingDetails) {
         this.trackingDetails = trackingDetails;
     }
 
+    // This method is a misspelling, but it persists to avoid breaking backwards compatibility
+
```
Justintime50, line 269: This comment needs to be tied to the `getUpdateAt` method and not free-floating as it is now.

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -717,15 +669,15 @@ public String getLabelUrl() {
...
-    // Tracker
-
     /**
      * @return the ID of this shipment
      */
     public String getShipmentId() {
         return "";
     }
 
+    // Batch
```
Justintime50, line 679: This comment was misplaced during the shuffle

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -740,11 +692,51 @@ public String getStatus() {
         return "";
     }
 
+    // Tracker
```
Justintime50, line 695: This comment was misplaced during the shuffle

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -740,11 +692,51 @@ public String getStatus() {
...
+
     /**
      * @return the tracking details of this shipment
      */
     public List<TrackingDetail> getTrackingDetails() {
         return new ArrayList<TrackingDetail>();
     }
 
+    protected enum RequestMethod {
+        GET, POST, DELETE, PUT
```
Justintime50, line 705: Comma separate lines read better one item per line and sorted, can we do that here?

src/test/java/com/easypost/EasyPostTest.java
```diff
@@ -1002,7 +1000,7 @@ public void testPaymentLogReport() throws EasyPostException {
     public void testWebhookCRUD() throws EasyPostException {
         // Define request params
         Map<String, Object> paramMap = new HashMap<String, Object>();
-        String url = "https://example.com/" + UUID.randomUUID().toString();
```
Justintime50, line 1005: Does `UUID.randomUUID()` already return a string?

src/test/java/com/easypost/EventTest.java
```diff
@@ -32,11 +32,7 @@ public void testEventWithNullPreviousAttributes() {
         eventWrapper.put("mode", "production");
         eventWrapper.put("previous_attributes", null);
         eventWrapper.put("pending_urls", new ArrayList<String>(
-                Arrays.asList(
-                        "hook_4ed7ff31c6974ff180e2463bb2912345",
-                        "hook_b43bc6d4eb3848bfaaee8e131f712345")
-                )
-        );
+                Arrays.asList("hook_4ed7ff31c6974ff180e2463bb2912345", "hook_b43bc6d4eb3848bfaaee8e131f712345")));
```
Justintime50, line 35: Same as above, I'd rather have comma separated lists be one item per line. Makes managing and scanning through them much easier.

Justintime50 (commented):
Couple real small things. Looks like a comment or two got misplaced during the shuffling.

src/main/java/com/easypost/net/EasyPostResource.java
nwithan8: We can, if I can find the .editorconfig setting to override enum values formatting

src/test/java/com/easypost/EasyPostTest.java
nwithan8: It returns a UUID object, which inherently knows how to convert to a string when added to another string. So the explicit conversion is unnecessary and the linter removed it.

src/main/java/com/easypost/net/EasyPostResource.java
Justintime50: Ok, don't spin wheels on it if it's not immediately obvious. Generally though that's the idea. One odd list won't break the world.

src/test/java/com/easypost/EventTest.java
nwithan8: We can do this manually, but the .editorconfig settings is going to keep reverting.
Here, `asList` is actually a function that can take an infinite number of arguments. But since it's a function, it adheres to the rules about how functions with parameters are formatted. So the rule to change would impact all functions.
Under the current rules, like any function, if the number of parameters exceeds the line length limit, it will kick it down to a second line.

src/main/java/com/easypost/net/EasyPostResource.java
nwithan8: Resolved in 28156240aa91125dfeeb163cd5eb6d71869b762a

src/main/java/com/easypost/net/EasyPostResource.java
nwithan8: Resolved in 28156240aa91125dfeeb163cd5eb6d71869b762a

src/main/java/com/easypost/net/EasyPostResource.java
nwithan8: Resolved in 28156240aa91125dfeeb163cd5eb6d71869b762a

src/main/java/com/easypost/model/Shipment.java
nwithan8: Resolved in 28156240aa91125dfeeb163cd5eb6d71869b762a

src/main/java/com/easypost/model/Tracker.java
nwithan8: Resolved in 28156240aa91125dfeeb163cd5eb6d71869b762a

src/test/java/com/easypost/EventTest.java
Justintime50: I see, this is a function (I hadn't realized) and not a free-form list. This is probably fine as is. Thanks for the info!

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -633,11 +525,72 @@ private static EasyPostResponse makeAppEngineRequest(final RequestMethod method,
...
-            new ArrayList<>(Arrays.asList("getCreatedAt", "getUpdatedAt", "getFees"));
-    private Date createdAt;
-    private Date updatedAt;
-    private ArrayList<Fee> fees;
+    /**
+     * Returns a string representation of the object.
+     */
+    @Override
+    public String toString() {
+
```
Justintime50, line 533: Nit: extra line here isn't necessary.

Justintime50 (approved):
Project linted, nothing was erroneously removed here.

jchen293 (approved):
lgtm

PR #93, opened 2022-01-28
Allow user override of default API timeouts

nwithan8:
- Add setters/getters for all three timeouts
- Add test to check that setters/getters operate as expected
- Code rearrangement per file layout rules


Justintime50:
Gonna wait for the linting task to land so this isn't so large or a PR containing things that aren't related.

Justintime50:
@nwithan8 can you rebase and I'll take another look?

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -46,54 +46,29 @@
...
-    }
+    public static final String CHARSET = "UTF-8";
+    public static final ArrayList<String> GLOBAL_FIELD_ACCESSORS =
+            new ArrayList<>(Arrays.asList("getCreatedAt", "getUpdatedAt", "getFees"));
+    private static final int DEFAULT_CONNECT_TIMEOUT_MILLISECONDS = 30000;
+    private static final int DEFAULT_READ_TIMEOUT_MILLISECONDS = 60000;
+    private static final double DEFAULT_APP_ENGINE_TIMEOUT_SECONDS = 20.0;
+    private static final String DNS_CACHE_TTL_PROPERTY_NAME = "networkaddress.cache.ttl";
+    // Set this property to override your environment's default URLStreamHandler.
+    private static final String CUSTOM_URL_STREAM_HANDLER_PROPERTY_NAME = "com.easypost.net.customURLStreamHandler";
```
jontsai, line 65: How would someone be able to set this property if it is declared as a `final` ?

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -633,11 +529,120 @@ private static EasyPostResponse makeAppEngineRequest(final RequestMethod method,
...
+     * Pretty print the JSON representation of the object.
+     *
+     * @return the JSON representation of the object.
+     */
+    public String prettyPrint() {
+        return String.format("<%s@%s id=%s> JSON: %s", this.getClass().getName(), System.identityHashCode(this),
+                this.getIdString(), PRETTY_PRINT_GSON.toJson(this));
+    }
+
+    private Object getIdString() {
```
jontsai, line 599: Should the return type be `String` ?

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -46,54 +46,29 @@
...
-    /**
-     * Returns a string representation of the object.
-     */
-    @Override
-    public String toString() {
-
-        return (String) this.getIdString();
-    }
-
-    /**
```
jontsai, line 73: # Reviewer Note

This block of code has been moved lower in the file

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -117,60 +92,15 @@ protected static String instanceURL(final Class<?> clazz, final String id) {
         return String.format("%s/%s", classURL(clazz), id);
     }
 
-    /**
```
jontsai, line 120: # Reviewer Note

This block of code has been moved lower in the file

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -343,40 +273,6 @@ private static String createQuery(final Map<String, Object> params) throws Unsup
         return flatParams;
     }
 
-    // represents Errors returned as JSON
```
jontsai, line 346: # Reviewer Note

This block of code has been moved lower in the file

src/main/java/com/easypost/net/EasyPostResource.java
Justintime50: Yeah, this PR still needs a rebase and most of these moving around bits will go away.

src/main/java/com/easypost/net/EasyPostResource.java
nwithan8: This property doesn't have a setter anywhere, and since it's private, wouldn't be able to be accessed directly.
Should we remove the comment or add in a way to set this?

src/test/java/com/easypost/EasyPostTest.java
```diff
@@ -124,7 +128,7 @@ public static void setUp() {
         canadaToAddress.put("zip", "t8n2m4");
         canadaToAddress.put("country", "CA");
     }
-
+
```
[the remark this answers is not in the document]
Justintime50: +1

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -703,7 +760,6 @@ public String getStatus() {
         PUT
     }
 
-    // represents Errors returned as JSON
```
Justintime50, line 706: Why was this removed? Why was it here to begin with?

nwithan8: I can re-add it if needed. I believe it was just an explainer comment on what the `Error` class was for.

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -59,10 +59,13 @@
             new ArrayList<>(Arrays.asList("getCreatedAt", "getUpdatedAt", "getFees"));
     private static final int DEFAULT_CONNECT_TIMEOUT_MILLISECONDS = 30000;
     private static final int DEFAULT_READ_TIMEOUT_MILLISECONDS = 60000;
```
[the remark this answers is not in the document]
nwithan8: That's what is was called originally, but we can change it.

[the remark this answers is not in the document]
Justintime50: What do you think makes most sense?

src/main/java/com/easypost/net/EasyPostResource.java
Justintime50: If we expect users would need/want to edit this, we should probably add getters/setters. If not, we could remove the comment. I frankly don't know off-hand what this is used for.

nwithan8: This has been a private variable with no setter/getter from the very beginning of the library (e0e912c2233d1720b065438e435ab6d4adcac93d). So, whatever it is, it's always been this way.
I vote for removing the comment.

Justintime50: Let's remove the comment. If someone really needs this bit we can always tweak it later.

nwithan8: Resolved in 4abf9d9

src/main/java/com/easypost/net/EasyPostResource.java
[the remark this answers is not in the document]
nwithan8: It's called `REQUEST` in our Java library, might as well be consistent.

Justintime50 (approved):

jchen293 (approved):
LGTM, you can also update the changelog for the feature that we added.

PR #96, opened 2022-01-31
Improve create list tracker

jchen293:
## What?
Improve the **createList** method in the Tracker class
## Why?
Currently, our **createList** method in the Tracker class is doing extra work which refactors the user input. The user is also creating an extra list and map. Overall, this extra work would downgrade the runtime performance which takes O(n) time complexity for **createList** method, and when the user mistypes "trackers" as the key for input map, a NullPointerException could be thrown.
## How?
Instead of asking the user to input a map that **must** have "**trackers**" as the key and a list of maps as value, we now just need the user to enter a map of maps where the key is the number of trackers and value is a map of **tracking_code** and **carrier**.
## Testing?
Modify the unit test to accommodate the code change.

### DIFF
--- a/src/main/java/com/easypost/model/Tracker.java
+++ b/src/main/java/com/easypost/model/Tracker.java
@@ -3,7 +3,6 @@ package com.easypost.model;
 import com.easypost.exception.EasyPostException;
 import com.easypost.net.EasyPostResource;
 
-import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


Justintime50 (approved):
Nice work!

nwithan8 (approved):

PR #98, opened 2022-02-02
support update brand.

jchen293:
# what
allow users to update the brand through API.

# why
Java library is currently missing this feature.

# how
Add Brand class and the updateBrand method in the User class.

# testing
E2E local test

### DIFF
--- a/src/main/java/com/easypost/model/Brand.java
+++ b/src/main/java/com/easypost/model/Brand.java
@@ -10,7 +10,7 @@ public class Brand extends EasyPostResource {
     private String ad;
     private String adHref;
     private String name;
-    private User userID;
+    private String userID;
     private String theme;
 
     /**
@@ -122,18 +122,18 @@ public class Brand extends EasyPostResource {
     }
 
     /**
-     * Get the name of the Brand.
+     * Get the name of the User.
      *
-     * @return the name of the Brand.
+     * @return the name of the User.
      */
     public String getName() {
         return this.name;
     }
 
     /**
-     * Set the name of the Brand.
+     * Set the name of the User.
      *
-     * @param name the name of the Brand.
+     * @param name the name of the User.
      */
     public void setName(String name) {
         this.name = name;
@@ -144,7 +144,7 @@ public class Brand extends EasyPostResource {
      *
      * @return the user ID of the Brand.
      */
-    public User getUserID() {
+    public String getUserID() {
         return this.userID;
     }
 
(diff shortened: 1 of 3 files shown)


nwithan8 (dismissed):
LGTM

src/main/java/com/easypost/model/Brand.java
```diff
@@ -0,0 +1,177 @@
...
+
+public class Brand extends EasyPostResource {
+    private String backgroundColor;
+    private String color;
+    private String logo;
+    private String logoHref;
+    private String ad;
+    private String adHref;
+    private String name;
+    private User userID;
```
Justintime50, line 13: This is a field and is a `String`

src/main/java/com/easypost/model/Brand.java
```diff
@@ -0,0 +1,177 @@
...
+     * Set the ad href of the Brand.
+     *
+     * @param adHref the ad href of the Brand.
+     */
+    public void setAdHref(String adHref) {
+        this.adHref = adHref;
+    }
+
+    /**
+     * Get the name of the Brand.
```
Justintime50, line 125:
```suggestion
     * Get the name of the User.
```

src/main/java/com/easypost/model/Brand.java
```diff
@@ -0,0 +1,177 @@
...
+     *
+     * @return the name of the Brand.
```
Justintime50, line 127:
```suggestion
     * @return the name of the User.
```

src/main/java/com/easypost/model/Brand.java
```diff
@@ -0,0 +1,177 @@
...
+     */
+    public String getName() {
+        return this.name;
+    }
+
+    /**
+     * Set the name of the Brand.
```
Justintime50, line 134:
```suggestion
     * Set the name of the User.
```

src/main/java/com/easypost/model/Brand.java
```diff
@@ -0,0 +1,177 @@
...
+     *
+     * @param name the name of the Brand.
```
Justintime50, line 136:
```suggestion
     * @param name the name of the User.
```

Justintime50 (changes requested):

src/main/java/com/easypost/model/Brand.java
jchen293: in our API doc, its User object: https://www.easypost.com/docs/api/node#brand

Justintime50: Although the user ID is the id of a `User` object, this is a field that expects/returns a string. Please update it. It should be `string` on our docs as well.

Justintime50 (approved):

nwithan8 (approved):

PR #99, opened 2022-02-03
Fix timeout values in test

nwithan8:
We had a bad testing setup, where the test for changing the API timeouts was running before other tests. Without resetting the timeouts, they were too aggressive and were causing other tests to fail.

This PR includes:
- Public (but immutable) default timeout values, so can be accessed if needed (i.e. to reset values)
- Corrected timeout test that resets what it changed after executing, so other tests can run correctly.

### DIFF
--- a/src/main/java/com/easypost/net/EasyPostResource.java
+++ b/src/main/java/com/easypost/net/EasyPostResource.java
@@ -546,13 +546,6 @@ public abstract class EasyPostResource {
     }
 
     /**
-     * Reset the API connection timeout to the default value.
-     */
-    public static void resetConnectTimeoutMilliseconds() {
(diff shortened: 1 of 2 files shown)


Justintime50:
This is interesting. If each of the tests are isolated and run separately, why did this even occur?

nwithan8:
> This is interesting. If each of the tests are isolated and run separately, why did this even occur?

The timeouts are static global variables, so once changed, they affect everything.

jchen293:
Can we put that API timeouts test in a separate file so it does not affect other unit tests?

Justintime50 (approved):

jchen293 (approved):

PR #100, opened 2022-02-04
remove duplicate code in lowestRate() method.

jchen293:
Remove duplicate code in the lowestRate() method. Current unit tests still passed.


Justintime50 (approved):
Oof, nice catch.

nwithan8 (approved):

PR #101, opened 2022-02-07
Get shipment rates

jchen293:
Currently, our Java library has a getter method **getRate()** but does not have a method to allow users to get rates if they create a shipment but don't buy it immediately. 

Add an additional method **retrieveRates()** that allows the user to get the latest rate of a specific shipment. Add unit test to ensure the code works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,9 +12,7 @@ Icon
 target
 
 .idea/
-src/main/main.iml
-src/test/test.iml
-easypost-api-client.iml
+*.iml
 
 *.class
 settings.xml
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ target
 .idea/
 src/main/main.iml
 src/test/test.iml
+easypost-api-client.iml
 
 *.class
 settings.xml
--- a/src/main/java/com/easypost/model/Shipment.java
+++ b/src/main/java/com/easypost/model/Shipment.java
@@ -1064,23 +1064,24 @@ public final class Shipment extends EasyPostResource {
     }
 
     /**
-     * Update the rates of a shipment.
+     * Get the rates of a shipment.
      *
+     * @return Rate object
      * @throws EasyPostException
      */
-    public void retrieveRates() throws EasyPostException {
-        this.retrieveRates(null);
+    public Shipment retrieveRates() throws EasyPostException {
+        return retrieveRates(null);
     }
 
     /**
-     * Update the rates of a shipment.
+     * Get the rates of a shipment.
      *
+     * @return Rate object
      * @param apiKey API key to use in request (overrides default API key).
      * @throws EasyPostException
      */
-    public void retrieveRates(String apiKey) throws EasyPostException {
-        Shipment response = request(RequestMethod.GET, String.format("%s/rates",
+    public Shipment retrieveRates(String apiKey) throws EasyPostException {
+        return request(RequestMethod.GET, String.format("%s/rates",
                         instanceURL(Shipment.class, this.getId())), null, Shipment.class, apiKey);
-        this.setRates(response.getRates());
     }
 }
--- a/src/main/java/com/easypost/model/Shipment.java
+++ b/src/main/java/com/easypost/model/Shipment.java
@@ -1062,4 +1062,29 @@ public final class Shipment extends EasyPostResource {
 
         return lowestRate;
     }
+
+    /**
+     * Get rate for a shipment.
+     *
+     * @return Shipment object
+     * @throws EasyPostException
+     */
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -14,6 +14,7 @@ target
 .idea/
 src/main/main.iml
 src/test/test.iml
+easypost-api-client.iml
```
Justintime50, line 17: Let's just wildcard this: `*.iml` and replace all three of these with it.

src/main/java/com/easypost/model/Shipment.java
```diff
@@ -1047,7 +1047,6 @@ public Rate lowestRate(final List<String> carriers, final List<String> services)
                 continue;
             }
             if (services != null && services.size() > 0 &&
-                    !services.contains(this.rates.get(i).getService().toLowerCase()) &&
```
Justintime50, line 1050: I thought we already removed this in separate PR?

src/main/java/com/easypost/model/Shipment.java
```diff
@@ -1063,4 +1062,28 @@ public Rate lowestRate(final List<String> carriers, final List<String> services)
...
         return lowestRate;
     }
+
+    /**
+     * Get rate for a shipment.
+     *
+     * @return Shipment object
+     * @throws EasyPostException
+     */
+    public Shipment getShipmentRate() throws EasyPostException {
```
Justintime50, line 1072: This doesn't just return one rate, it returns a list of rates so it should be pluralized. This is true of all the `getShipmentRate` occurrences. I'm also torn on the name of this. In our other libraries we call it `getRates` but we obviously have the getter already titled `getRates`... I need to think on this for a second.

Justintime50 (dismissed):

src/main/java/com/easypost/model/Shipment.java
jchen293: forgot to rebase, it's resolved.

nwithan8 (approved):

src/main/java/com/easypost/model/Shipment.java
```diff
@@ -1062,4 +1062,28 @@ public Rate lowestRate(final List<String> carriers, final List<String> services)
...
+     *
+     * @return Shipment object
+     * @throws EasyPostException
+     */
+    public Shipment retrieveRates() throws EasyPostException {
+        return retrieveRates(null);
+    }
+
+    /**
+     * Get rate for a shipment.
```
Justintime50, line 1077:
```suggestion
     * Get the rates of a shipment.
```

src/main/java/com/easypost/model/Shipment.java
```diff
@@ -1062,4 +1062,28 @@ public Rate lowestRate(final List<String> carriers, final List<String> services)
...
+     *
+     * @param apiKey API key to use in request (overrides default API key).
+     * @return Shipment object
+     * @throws EasyPostException
+     */
+    public Shipment retrieveRates(String apiKey) throws EasyPostException {
+        Shipment response = request(RequestMethod.GET, String.format("%s/rates",
+                        instanceURL(Shipment.class, this.getId())), null, Shipment.class, apiKey);
+        this.setRates(response.getRates());
```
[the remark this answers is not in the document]
jchen293: Just in case the user is still using the old shipment object, including the setter here will update the old object's rates to the latest rates.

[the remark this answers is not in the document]
jchen293: The method is changed to void, we are updating the current object's rates instead of returning a new shipment object.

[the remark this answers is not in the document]
Justintime50: This function does not update the rates, it only retrieves the rates. Let's remove this set line here.

src/main/java/com/easypost/model/Shipment.java
```diff
@@ -1062,4 +1062,26 @@ public Rate lowestRate(final List<String> carriers, final List<String> services)
...
+    }
+
+    /**
+     * Get the rates of a shipment.
+     *
+     * @return Rate object
+     * @param apiKey API key to use in request (overrides default API key).
+     * @throws EasyPostException
+     */
+    public Shipment retrieveRates(String apiKey) throws EasyPostException {
```
Justintime50, line 1083: Thinking on the usage of this function tonight. Will get back to you tomorrow (static vs non-static - it may need to be adjusted across all our libraries).

Justintime50 (changes requested):

Justintime50:
We've determined that we are not going to support the `shipments/{id}/rates` endpoint in the CLs and this is not needed. If you needed the rates of a shipment, simply retrieve the shipment and the rates will be present.

PR #103, opened 2022-02-16
add full test coverage.

jchen293:
This PR adds full test coverage to the library by testing everything "reasonable". We didn't test every getter/setter in the test suite. Squashed a bug and add missing getters/setters.

- Convert all the tests in EasyPostTest.java to individual files corresponding to their class. 
- Add more unit tests that are not covered in EasyPostTest.java.
- Created a utility class called Fixture that allowed us to reuse the code.
- Add missing getter/setter for eel_pfc in CustomsInfo class.
- Add missing getter/setter for id in Brand class.
- Fix create() method in Refund class.
- Upgrade junit4 to junit5
- Add [assertj](https://assertj.github.io/doc/#overview-what-is-assertj) dependency to compare two objects in the unit test
- Fix typo in AddressCollection class, change Batch to Address.
- Add CI test in Github Actions

Screen Shot 2022-02-23 at 2 29 31 PM


easypost-api-client.iml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Justintime50, line 1: What is this file for?

src/test/java/com/easypost/AddressTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+public class AddressTest {
+
```
Justintime50, line 21: No line break between the beginning of a class and its code.

src/test/java/com/easypost/AddressTest.java
```diff
@@ -0,0 +1,101 @@
...
+    public void testCreate() throws EasyPostException {
+        Address address = Address.create(Fixture.basicAddress());
+
+        assertNotNull(address);
+        assertEquals("388 Townsend St", address.getStreet1());
+        assertTrue(address.getId().startsWith("adr_"));
+    }
+
+    @Test
+    public void testCreateVerifiyStrict() throws EasyPostException {
```
Justintime50, line 40:
```suggestion
    public void testCreateVerifyStrict() throws EasyPostException {
```

src/test/java/com/easypost/AddressTest.java
```diff
@@ -0,0 +1,101 @@
...
+        Map<String, Object> addressData = Fixture.basicAddress();
+        List<String> verificationList = new ArrayList<>();
+        verificationList.add("true");
+        addressData.put("verify_strict", verificationList);
+
+        Address address = Address.create(addressData);
+
+        assertNotNull(address);
```
Justintime50, line 48: Please follow the PHP library and assert that these are the correct instances. Asserting that an object is not null is not a good test when we immediately follow it up with asserting an exact value (such as street1 on the next line), because then obviously if that's true, the object couldn't have been null. Let's keep these as close to the PHP one as possible.

src/test/java/com/easypost/AddressTest.java
```diff
@@ -0,0 +1,101 @@
...
+        assertEquals("388 TOWNSEND ST APT 20", address.getStreet1());
+        assertTrue(address.getId().startsWith("adr_"));
+    }
+
+    @Test
+    public void testRetrieve() throws EasyPostException{
+        Address retrievedAddress = Address.retrieve(address.getId());
+
+        assertNotNull(retrievedAddress);
```
Justintime50, line 57: Same comment as above on asserting the instance and following the PHP lib.

src/test/java/com/easypost/AddressTest.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Test
+    public void testAll() throws EasyPostException{
+        Map<String, Object> params = new HashMap<>();
+        params.put("page_size", Fixture.pageSize());
+
+        AddressCollection addresses = Address.all(params);
+        List<Batch> addressesArray = addresses.getAddresses();
+        assertNotNull(addressesArray);
```
Justintime50, line 68: Please split assertions and the logic to create/retrieve records with a line break. This makes it really easy for the eye to skim through these and get the info you need. This is true of all assertions everywhere. eg:

```
# Block of code creating stuff
1...
2...

# Block of code asserting stuff
1...
2...
```

src/test/java/com/easypost/AddressTest.java
```diff
@@ -0,0 +1,101 @@
...
+        for(Batch address: addressesArray) {
+            assertTrue(address.getId().startsWith("adr_"));
+        }
+    }
+
+    @Test
+    public void testCreateVerify() throws EasyPostException {
+        Address address = Address.create(Fixture.incorrectAddressToVerify());
+
+        assertNotNull(address);
```
Justintime50, line 79: Same comment as above on asserting the instance and following the PHP lib.

src/test/java/com/easypost/AddressTest.java
```diff
@@ -0,0 +1,101 @@
...
+        assertEquals("417 MONTGOMERY ST STE 500", address.getStreet1());
+        assertTrue(address.getId().startsWith("adr_"));
+    }
+
+    @Test
+    public void testCreateAndVerify() throws EasyPostException {
+        Address address = Address.createAndVerify(Fixture.incorrectAddressToVerify());
+
+        assertNotNull(address);
```
Justintime50, line 88: Same comment as above on asserting the instance and following the PHP lib.

src/test/java/com/easypost/AddressTest.java
```diff
@@ -0,0 +1,101 @@
...
+        assertEquals("417 MONTGOMERY ST STE 500", address.getStreet1());
+        assertTrue(address.getId().startsWith("adr_"));
+    }
+
+    @Test
+    public void testVerify() throws EasyPostException {
+        address.verify();
+
+        assertNotNull(address);
```
Justintime50, line 97: Same comment as above on asserting the instance and following the PHP lib.

src/test/java/com/easypost/Fixture.java
```diff
@@ -0,0 +1,175 @@
...
+import java.util.Map;
+
+public class Fixture {
+
+    public static int pageSize() {
+        return 5;
+    }
+
+    public static String uspsCarrierAccountID() {
+        return "ca_8dc116debcdb49b5a66a2ddee4612600";
```
Justintime50, line 15: This needs to be the carrier account ID of the USPS carrier account for the Java user. This is the one for the PHP user.

src/test/java/com/easypost/Fixture.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    public static String childUserID() {
+        return "user_3878404c0bdd4321952d4cae45c1d184";
```
Justintime50, line 19: This needs to be the user ID of a child user for the Java user. This is the one for the PHP user.

src/test/java/com/easypost/Fixture.java
```diff
@@ -0,0 +1,175 @@
...
+    public static String usps() {
+        return "USPS";
+    }
+
+    public static String uspsService() {
+        return "First";
+    }
+
+    public static String reportStartDate() {
+        return "2022-02-01";
```
Justintime50, line 31: Once you have created a few records, you'll most likely need these to be todays date. The Java user probably doesn't have data for this date range.

Justintime50 (changes requested):
You're almost on the right path. I reviewed the `Address` object tests. Please take those comments and apply them throughout the whole PR and moving forward.

easypost-api-client.iml
jchen293: It's a module settings file created by IntelliJ IDEA, I will add it to the **gitingore** file

src/main/java/com/easypost/model/CustomsInfo.java
```diff
@@ -228,4 +229,22 @@ public void setRestrictionComments(final String restrictionComments) {
     public void setCustomsItems(final List<CustomsItem> customsItems) {
         this.customsItems = customsItems;
     }
+
+    /**
+     * Get this CustomsInfo's eel pfc.
+     *
+     * @return the eel pfc of this CustomsInfo.
```
Justintime50, line 236: These are acronyms:

```suggestion
     * @return the EEL or PFC of this CustomsInfo.
```

src/main/java/com/easypost/model/CustomsInfo.java
```diff
@@ -228,4 +229,22 @@ public void setRestrictionComments(final String restrictionComments) {
...
+     */
+    public String getEelPfc() {
+        return eelPfc;
+    }
+
+    /**
+     * Set this CustomsInfo's eel pfc.
+     *
+     * @param eelPfc the eel pfc of this CustomsInfo.
```
Justintime50, line 245: These are acronyms:

```suggestion
     * @param eelPfc the EEL or PFC of this CustomsInfo.
```

src/main/java/com/easypost/model/Refund.java
```diff
@@ -38,7 +39,8 @@
         Map<String, Object> wrappedParams = new HashMap<String, Object>();
         wrappedParams.put("refund", params);
 
-        return request(RequestMethod.POST, classURL(Refund.class), wrappedParams, List.class, apiKey);
+        Refund[] response = request(RequestMethod.POST, classURL(Refund.class), wrappedParams, Refund[].class, apiKey);
```
Justintime50, line 42: This isn't correct. The response of a Refund should be a refreshed Shipment object, not an array of refunds to list. The `refund_status` would eventually say `refunded`.

src/test/java/com/easypost/AddressTest.java
```diff
@@ -0,0 +1,145 @@
...
+     *
+     *  @throws EasyPostException when the request fails.
+     */
+    @Test
+    public void testCreateVerify() throws EasyPostException {
+        Address address = Address.create(Fixture.incorrectAddressToVerify());
+
+        assertTrue(address instanceof Address);
+        assertEquals("417 MONTGOMERY ST STE 500", address.getStreet1());
+        assertTrue(address.getId().startsWith("adr_"));
```
Justintime50, line 114: Let's try and keep all the assertions in the same order (1. instance, 2. ID, 3. property).

This goes for all tests in the library.

src/test/java/com/easypost/ShipmentTest.java
```diff
@@ -0,0 +1,224 @@
...
+     * @throws EasyPostException when the request fails.
+     */
+    @Test
+    public void testCreate() throws EasyPostException {
+        Shipment shipment = Shipment.create(Fixture.fullShipment());
+
+        assertTrue(shipment instanceof Shipment);
+        assertTrue(shipment.getId().startsWith("shp_"));
+        assertEquals("PNG", shipment.getOptions().get("label_format"));
+        assertEquals(123.0, shipment.getOptions().get("invoice_number"));
```
Justintime50, line 48: This shouldn't be right. `123` does not equal `123.0`.

src/test/java/com/easypost/ShipmentTest.java
```diff
@@ -0,0 +1,224 @@
...
+    /**
+     * Test getting smart rate from a shipment.
+     *
+     * @throws EasyPostException when the request fails.
+     */
+    @Test
+    public void testSmartRate() throws EasyPostException {
+        List<Rate> smartRates = shipment.getSmartrates();
+
+        assertTrue(smartRates instanceof List);
```
Justintime50, line 181: Please write this test like it's written in the other libraries. Asserting that each field is filled. We wrote it this way previously because we had initial issues on those fields being null.

src/test/java/com/easypost/UserTest.java
```diff
@@ -1,64 +1,141 @@
...
+     * @throws EasyPostException when the request fails.
+     */
+    @BeforeClass
+    public static void setUp() throws EasyPostException{
+        EasyPost.apiKey = "EASYPOST_PROD_API_KEY";
+        user = User.retrieveMe();
+    }
+
+    /**
+     * Test creating a user.
```
Justintime50, line 32: This should say `child user`
```suggestion
     * Test creating a child user.
```

src/main/java/com/easypost/model/Brand.java
```diff
@@ -12,6 +12,7 @@
     private String name;
     private String userID;
     private String theme;
+    private String id;
```
Justintime50, line 15: Please write up a ticket internally (Phab) to document the `id` field in our docs, we'll do that separate.

src/test/java/com/easypost/ShipmentTest.java
jchen293: getOptions() returns a Hashmap, the value for "invoice_number" key is double which is 123.0

src/main/java/com/easypost/model/Refund.java
jchen293: This is **create()** method in the Refund class, I think you are talking about **refund()** method in the Shipment class.

src/test/java/com/easypost/ShipmentTest.java
Justintime50: That's definitely wrong then. The type of this field is a string. Please correct it from `double` to `string` and add a line entry to the CHANGELOG for the `NEXT RELEASE` saying this was corrected.

src/main/java/com/easypost/model/Refund.java
Justintime50: Ah, another janky endpoint. You are correct... it returns a list of JSON objects. This is fine then.

src/test/java/com/easypost/ShipmentTest.java
jchen293: **`invoice_number`** does not belong to any class attribute, apparently, it was because the test input is double, I have changed the type in Fixture from `double` to `string`.

src/test/java/com/easypost/AddressTest.java
```diff
@@ -0,0 +1,152 @@
...
+     *
+     * @throws EasyPostException when the request fails.
+     */
+    @Test
+    public void testRetrieve() throws EasyPostException{
+        Address retrievedAddress = Address.retrieve(address.getId());
+
+        assertTrue(retrievedAddress instanceof Address);
+        assertTrue(retrievedAddress.getId().startsWith("adr_"));
+        assertEquals(address.getId(), retrievedAddress.getId());
```
Justintime50, line 78: We should be asserting the entire retrieved object matches the entire created object instead of just the IDs. There's only a few objects where asserting the ID only makes sense (batches for instance because the object changes between requests as it moves through the queue and processes the batch).

Please go through the entire test suite and have the assertions match what the PHP library is as close as possible. If there is a reason we cannot assert against the entire object, call it out here so I'm made aware of it.

src/test/java/com/easypost/AddressTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+        assertTrue(address instanceof Address);
+        assertTrue(address.getId().startsWith("adr_"));
+        assertEquals("417 MONTGOMERY ST STE 500", address.getStreet1());
+        assertEquals("CA", address.getState());
+        assertEquals("SAN FRANCISCO", address.getCity());
+        assertEquals("94104-1100", address.getZip());
+        assertEquals("4151234567", address.getPhone());
+        assertEquals("EASYPOST", address.getCompany());
+        assertEquals("US", address.getCountry());
```
Justintime50, line 136: Please follow the PHP library - all these additional assertions are unnecessary and complicate the test. `street1` is sufficient here as it corrects, thus we can infer that the rest of the address was also corrected.

After chatting, this was done to increase coverage (thank you!); however, for simplicity, let's stick to the example and remove all the additional assertions for now - we may just configure the coverage platform to ignore getters/setters.

src/test/java/com/easypost/BatchTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+        List<Shipment> shipmentData = new ArrayList<>();
+        Map<String, Object> params = new HashMap<>();
+
+        shipmentData.add(shipment);
+        params.put("shipments", shipmentData);
+
+        Batch batchWithAddedShipment = batch.addShipments(params);
+
+        assertTrue(batchWithAddedShipment.getId().startsWith("batch_"));
```
Justintime50, line 159: Please follow the PHP library's test suite. Asserting the ID here isn't necessary as we aren't doing anything related to the ID - we are only adding and removing something from the object and these assertions are handled elsewhere in other tests.

src/test/java/com/easypost/BatchTest.java
```diff
@@ -0,0 +1,199 @@
...
+        assertEquals(0, batchWithoutShipment.getNumShipments().intValue());
+    }
+
+    /**
+     * Test generating a label for a Batch.
+     *
+     * @throws EasyPostException when the request fails.
+     */
+    @Test
+    @Disabled // This test is skipped because we have to wait the batch status is created and postage_purchased.
```
Justintime50, line 174:
```suggestion
    @Disabled // This test is skipped because we have to wait for the batch status to be `created` and the shipment's status to be `postage_purchased`.
```

src/test/java/com/easypost/CarrierAccountTest.java
```diff
@@ -0,0 +1,135 @@
...
+     * Test retrieving all carrier accounts.
+     *
+     * @throws EasyPostException when the request fails.
+     */
+    @Test
+    @Order(3)
+    public void testAll() throws EasyPostException {
+        List<CarrierAccount> carrierAccounts = CarrierAccount.all(null);
+
+        assertTrue(carrierAccounts != null);
```
Justintime50, line 82: Let's remove this because the following line will already be testing if the `carrierAccounts` is not null.

src/test/java/com/easypost/CarrierAccountTest.java
```diff
@@ -0,0 +1,135 @@
...
+        assertTrue(carrierAccount instanceof CarrierAccount);
+    }
+
+    /**
+     * Test retrieving the carrier account types available.
+     *
+     * @throws EasyPostException when the request fails.
+     */
+    @Test
+    @Order(6)
```
Justintime50, line 129: The order of this one doesn't matter and it can be removed.

src/test/java/com/easypost/CustomsInfoTest.java
```diff
@@ -0,0 +1,51 @@
...
+     * Test retrieving a CustomsInfo.
+     *
+     * @throws EasyPostException when the request fails.
+     */
+    @Test
+    public void testRetrieve() throws EasyPostException {
+        CustomsInfo retrievedCustomsInfo = CustomsInfo.retrieve(customsInfo.getId());
+
+        assertTrue(retrievedCustomsInfo instanceof CustomsInfo);
+        assertEquals(customsInfo.getId(), retrievedCustomsInfo.getId());
```
Justintime50, line 49: Let's assert the entire object (this goes for all retrieve tests for the whole library)

src/test/java/com/easypost/InsuranceTest.java
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * Test creating an insurance object.
+     *
+     * @throws EasyPostException when the request fails.
+     */
+    @Test
+    public void testCreate() throws EasyPostException {
+        Insurance insurance = Insurance.create(params);
+
+        assertTrue(insurance instanceof Insurance);
```
Justintime50, line 48: Please add an assertion to check the amount field and make sure it matches (sorry I didn't do this in the PHP lib)

src/test/java/com/easypost/PickupTest.java
```diff
@@ -0,0 +1,99 @@
...
+        assertTrue(retrievedPickup.getId().startsWith("pickup_"));
+    }
+
+    /**
+     * Test buying a pickup.
+     *
+     * @throws EasyPostException when the request fails.
+     */
+    @Test
+    @Disabled // This test is ignore for now because it relies on the pickup min and max date
```
Justintime50, line 70:
```suggestion
    @Disabled // This test is ignored for now because it relies on the pickup min and max date
```

Justintime50: Let's add a TODO note that we'll revisit this once we have VCR

src/test/java/com/easypost/PickupTest.java
```diff
@@ -0,0 +1,99 @@
...
+        assertEquals("scheduled", boughtPickup.getStatus());
+    }
+
+    /**
+     * Test cancelling a pickup.
+     *
+     * @throws EasyPostException when the request fails.
+     */
+    @Test
+    @Disabled // This test is ignore for now because it relies on the pickup min and max date
```
Justintime50, line 91:
```suggestion
    @Disabled // This test is ignored for now because it relies on the pickup min and max date
```

Justintime50: Let's add a TODO note to fix this when we have VCR

src/test/java/com/easypost/ShipmentTest.java
```diff
@@ -0,0 +1,234 @@
...
+    @Test
+    public void testAll() throws EasyPostException {
+        Map<String, Object> params = new HashMap();
+
+        params.put("page_size", Fixture.pageSize());
+
+        ShipmentCollection shipments = Shipment.all(params);
+
+        assertNotNull(shipments.getHasMore());
+        assertTrue(shipments.getShipments().size() <=5);
```
Justintime50, line 77: Nit: 
```suggestion
        assertTrue(shipments.getShipments().size() <= 5);
```

src/test/java/com/easypost/ShipmentTest.java
```diff
@@ -0,0 +1,234 @@
...
+     *
+     * @throws EasyPostException when the request fails.
+     */
+    @Test
+    public void testInsure() throws EasyPostException {
+        Map<String, Object> shipmentData = Fixture.oneCallBuyShipment();
+        Map<String, Object> insuranceData = new HashMap<>();
+
+        insuranceData.put("amount", "100");
+        shipmentData.put("insurance", 0);
```
Justintime50, line 147: Please add the following inline comment here: `Set to 0 so USPS doesn't insure this automatically and we can insure the shipment manually`

This explains they **why** we are setting it to 0 which is super important to call out.

src/test/java/com/easypost/ShipmentTest.java
```diff
@@ -0,0 +1,234 @@
...
+        Shipment shipmentWithInsurance = shipment.insure(insuranceData);
+
+        assertTrue(shipmentWithInsurance instanceof Shipment);
+        assertTrue(shipmentWithInsurance.getId().startsWith("shp_"));
+        assertEquals("100.00", shipmentWithInsurance.getInsurance());
+    }
+
+    /**
+     * Test refunding a shipment.
+     *
```
Justintime50, line 160: Please add the following to the docstring:

```
     * Refunding a test shipment must happen within seconds of the shipment being created as test shipments naturally
     * follow a flow of created -> delivered to cycle through tracking events in test mode - as such anything older
     * than a few seconds in test mode may not be refundable.
```

src/test/java/com/easypost/ShipmentTest.java
```diff
@@ -0,0 +1,234 @@
...
+        shipmentData.put("options", null);
+        shipmentData.put("tax_identifiers", null);
+        shipmentData.put("reference", "");
+
+        Shipment shipment = Shipment.create(shipmentData);
+
+        assertTrue(shipment instanceof Shipment);
+        assertTrue(shipment.getId().startsWith("shp_"));
+        assertNull(shipment.getReference());
+        assertNotNull(shipment.getOptions());
```
Justintime50, line 212: Please add inline comment `The EasyPost API populates some default values here`

src/test/java/com/easypost/ShipmentTest.java
```diff
@@ -0,0 +1,234 @@
...
+        Shipment shipmentWithLabel = shipment.label(params);
+
+        assertTrue(shipmentWithLabel instanceof Shipment);
+        assertTrue(shipmentWithLabel.getId().startsWith("shp_"));
+        assertNotNull(shipmentWithLabel.getPostageLabel().getLabelZplUrl());
+    }
+
+    /**
+     * Test adding insurace to a shipment.
+     *
```
Justintime50, line 138: Please add the following to the docstring:

```
     * If the shipment was purchased with a USPS rate, it must have had its insurance set to `0` when bought
     * so that USPS doesn't automatically insure it so we could manually insure it here.
```

src/test/java/com/easypost/WebhookTest.java
```diff
@@ -0,0 +1,113 @@
...
+        Map<String, Object> p = new HashMap<>();
+        p.put("url", "http://examples.com");
+
+        Webhook webhook = Webhook.create(p);
+
+        assertTrue(webhook instanceof Webhook);
+        assertTrue(webhook.getId().startsWith("hook_"));
+        assertEquals("http://examples.com", webhook.getUrl());
+
+        webhook.delete();
```
Justintime50, line 54: Please add an inline comment stating we are deleting the webhook here so we don't keep sending events to a dead webhook.

src/test/java/com/easypost/WebhookTest.java
```diff
@@ -0,0 +1,113 @@
...
+     *
+     * @throws EasyPostException when the request fails.
+     */
+    @BeforeAll
+    public static void setup() throws EasyPostException {
+        EasyPost.apiKey = System.getenv("EASYPOST_TEST_API_KEY");
+
+        params.put("url", "http://example.com");
+
+        webhook = Webhook.create(params);
```
Justintime50, line 34: Let's go ahead and prepend `global` to each of these objects created in the `setup` function. 

Let's do this throughout the whole project's test suite.

Justintime50 (changes requested):
Once you address the previous batch, I'll approve.

src/test/java/com/easypost/Fixture.java
```diff
@@ -0,0 +1,281 @@
...
+    /**
+     * Shipment that has all detail and is reusable in all test coverage.
+     *
+     * @return A map that have the shipment info filled in detail.
+     */
+    public static Map<String, Object> fullShipment() {
+        Map<String, Object> fullShipment = new HashMap<>();
+        Map<String, Object> options = new HashMap<>();
+
+        options.put("label_format", "PNG");
```
Justintime50, line 234: The line comment `Must be PNG so we can convert to ZPL later` is missing here

src/test/java/com/easypost/Fixture.java
```diff
@@ -0,0 +1,281 @@
...
+        options.put("invoice_number", "123");
```
Justintime50, line 235: What happens when you make this field an int? It probably complains because it's strongly typed eh?

src/test/java/com/easypost/Fixture.java
```diff
@@ -0,0 +1,281 @@
...
+        fullShipment.put("to_address", basicAddress());
+        fullShipment.put("from_address", basicAddress());
+        fullShipment.put("parcel", basicParcel());
+        fullShipment.put("customs_info", basicCustomsInfo());
+        fullShipment.put("reference", "123");
```
Justintime50, line 240: What happens when you make this field an int? It probably complains because it's strongly typed eh?

src/test/java/com/easypost/Fixture.java
jchen293: It still going to work. In the unit test assertions, we have to change the expected result to int.

src/test/java/com/easypost/Fixture.java
jchen293: The `reference` attribute in shipment is String, so yes it's strongly typed.

src/test/java/com/easypost/Fixture.java
jchen293: Nice catch.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # CHANGELOG
 
+## Next Release
+
+* Add getter and setter for eel_pfc in CustomsInfo class
+* Add getter and setter for id in Brand class
+* Fix bug for `create()` method in Refund class
+* Add test suite and upgrade JUnit4 to JUnit5
+* Add assertj dependency to compare two objects in unit test
```
Justintime50, line 9: This line isn't necessary to callout since it doesn't effect the user. The CHANGELOG is typically reserved for big highlights that change behavior of the app, but normal users will never run this test suite.

pom.xml
```diff
@@ -42,9 +42,16 @@
             <version>2.8.9</version>
         </dependency>
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>4.13.2</version>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <!-- use 2.9.1 for Java 7 projects -->
```
Justintime50, line 47: This line shouldn't be included since we don't even support Java 7.

src/main/java/com/easypost/model/AddressCollection.java
```diff
@@ -5,15 +5,15 @@
 import java.util.List;
 
 public final class AddressCollection extends EasyPostResource {
-    private List<Batch> addresses;
+    private List<Address> addresses;
```
Justintime50, line 8: This fix needs to be listed in the CHANGELOG

src/test/java/com/easypost/BatchTest.java
```diff
@@ -0,0 +1,197 @@
...
+        Map<String, Object> params = new HashMap<>();
+        
+        params.put("page_size", Fixture.pageSize());
+
+        BatchCollection batches = Batch.all(params);
+
+        assertTrue(batches.getBatches().size() <= Fixture.pageSize());
+        assertNotNull(batches.getHasMore());
+        for(Batch batch: batches.getBatches()) {
+            assertTrue(batch.getId().startsWith("batch_"));
```
Justintime50, line 82: This should really be checking if each object is an instance of batch.

src/test/java/com/easypost/AddressTest.java
```diff
@@ -0,0 +1,145 @@
...
+        params.put("page_size", Fixture.pageSize());
+
+        AddressCollection addresses = Address.all(params);
+
+        List<Address> addressesArray = addresses.getAddresses();
+
+        assertTrue(addressesArray.size() <= Fixture.pageSize());
+        assertNotNull(addresses.getHasMore());
+        for(Address address: addressesArray) {
+            assertTrue(address.getId().startsWith("adr_"));
```
Justintime50, line 98: This should really be checking if each item is an instance of address.

src/test/java/com/easypost/CarrierAccountTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+    /**
+     * Test creating a carrier account.
+     *
+     * @throws EasyPostException when the request fails.
+     */
+    @Test
+    @Order(1)
+    public void testCreate() throws EasyPostException {
+        CarrierAccount carrierAccount = CarrierAccount.create(params);
```
Justintime50, line 53: This carrier account needs to be deleted once it's done being tested.

src/test/java/com/easypost/EventTest.java
```diff
@@ -1,47 +1,61 @@
...
+     * @throws EasyPostException when the request fails.
+     */
+    @Test
+    public void testAll() throws EasyPostException {
+        EventCollection events = Event.all(params);
+
+        assertTrue(events.getEvents().size() <= Fixture.pageSize());
+        assertNotNull(events.getHasMore());
+        for(Event event: events.getEvents()) {
+            assertTrue(event.getId().startsWith("evt_"));
```
Justintime50, line 44: This should really assert each item is an instance of Event - this is true for all the `all` tests, they should all be adjusted accordingly throughout the suite.

src/test/java/com/easypost/Fixture.java
Justintime50: I ask because PHP for instance will lazily let an integer through and convert it on the way to string before sending it to the API; however, Ruby won't. I had to adjust them to strings in Ruby. The reason they are ints in the PHP lib is to test our serialization, that's why those comments were there. As this is a strongly typed language, let's keep these as they are.

src/test/java/com/easypost/Fixture.java
Justintime50: I ask because PHP for instance will lazily let an integer through and convert it on the way to string before sending it to the API; however, Ruby won't. I had to adjust them to strings in Ruby. The reason they are ints in the PHP lib is to test our serialization, that's why those comments were there. As this is a strongly typed language, let's keep these as they are.

src/test/java/com/easypost/InsuranceTest.java
```diff
@@ -0,0 +1,88 @@
...
+        params.put("page_size", Fixture.pageSize());
+
+        InsuranceCollection insurances = Insurance.all(params);
+
+        assertTrue(insurances instanceof InsuranceCollection);
+        assertNotNull(insurances.getHasMore());
+        assertTrue(insurances.getInsurances().size() <= Fixture.pageSize());
+        for(Insurance insurance: insurances.getInsurances()) {
+            assertTrue(insurance instanceof Insurance);
+            assertTrue(insurance.getId().startsWith("ins_"));
```
Justintime50, line 85: We can assert the instance only here, both aren't necessary

src/test/java/com/easypost/OrderTest.java
```diff
@@ -0,0 +1,106 @@
...
+        params.put("carrier", Fixture.usps());
+        params.put("service", Fixture.uspsService());
+
+        globalOrder.buy(params);
+
+        List<Shipment> shipments = globalOrder.getShipments();
+
+        assertTrue(globalOrder instanceof Order);
+        for(Shipment shipment: shipments) {
+            assertTrue(shipment.getId().startsWith("shp_"));
```
Justintime50, line 102: Not necessary

src/test/java/com/easypost/WebhookTest.java
Justintime50: Inline comments are "inline" because they either happen right before the line they are referencing or are on the same line. Stating the comment after the code it's annotating can be misleading. Please move the comment up a line.

Justintime50 (changes requested):

nwithan8 (dismissed):

.github/workflows/ci.yml
```diff
@@ -6,7 +6,7 @@ on:
   pull_request: ~
 
 jobs:
-  build:
+  build and test:
```
Justintime50, line 9: Please keep this name as `build`, changing the name will not allow this to get merged because the settings are looking for a `build` step to pass.

You also can't give names spaces.

.github/workflows/ci.yml
```diff
@@ -19,6 +19,8 @@ jobs:
           java-version: ${{ matrix.javaversion }}
       - name: Build with Maven
         run: mvn --batch-mode install -Dgpg.skip=true -DskipTests=true -Dcheckstyle.skip=true
+      - name: Test with Maven
+        run: mvn test -B
```
Justintime50, line 23:
```suggestion
        run: mvn --batch-mode install -Dgpg.skip=true -Dcheckstyle.skip=true
```

Justintime50: Instead of having an explicit step, we can simply remove the flag that skips tests as a part of the build process since by default in Java it will run both.

Justintime50 (changes requested):

Justintime50 (approved):
Nice work!

nwithan8 (approved):

PR #107, opened 2022-03-12
chore: add shipment ID test

Justintime50:
Adds a test when only IDs are passed to object creation. This was brought up as a way to check behavior across all libs due to: https://github.com/EasyPost/easypost-node/issues/222

### DIFF
--- a/src/test/java/com/easypost/ShipmentTest.java
+++ b/src/test/java/com/easypost/ShipmentTest.java
@@ -80,10 +80,7 @@ public class ShipmentTest {
 
         assertTrue(shipmentsList.size() <= Fixture.pageSize());
         assertNotNull(shipments.getHasMore());
-        assertTrue(shipments.getShipments().size() <= Fixture.pageSize());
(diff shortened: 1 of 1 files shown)


src/test/java/com/easypost/ShipmentTest.java
```diff
@@ -77,7 +80,10 @@ public void testAll() throws EasyPostException {
 
         assertTrue(shipmentsList.size() <= Fixture.pageSize());
         assertNotNull(shipments.getHasMore());
-        assertTrue(shipmentsList.stream().allMatch(shipment -> shipment instanceof Shipment));
```
jchen293, line 80: Was this accidentally changed?

jchen293 (approved):

src/test/java/com/easypost/ShipmentTest.java
Justintime50: Huh... not sure where how this came about, I'll patch this.

jchen293 (approved):

nwithan8 (approved):

PR #108, opened 2022-03-21
Support `columns` and `additional_columns` for reports

jchen293:
This PR adds support for specifying `columns` and `additional_columns` in the params for the report create method.

Adds two unit tests for this feature. E2E test and download the CSV report to check if the result is correct.

Additional columns and columns in the param, total three columns (from_name, from_company, usps_zone)
Screen Shot 2022-03-21 at 2 13 20 PM

Only columns in the param, only one column (usps_zone)
Screen Shot 2022-03-21 at 2 12 56 PM

### DIFF
--- a/src/main/java/com/easypost/model/Report.java
+++ b/src/main/java/com/easypost/model/Report.java
@@ -8,6 +8,8 @@ import java.net.URLEncoder;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.List;
+import java.util.ArrayList;
 
 public final class Report extends EasyPostResource {
     private String id;
@@ -39,13 +41,33 @@ public final class Report extends EasyPostResource {
      * @throws EasyPostException when the request fails.
      */
     public static Report create(final Map<String, Object> params, final String apiKey) throws EasyPostException {
+        String type = (String) params.get("type");
+        String url = reportURL(type) + "?";
+
+        if (params.containsKey("columns")) {
+            List<String> columns = (ArrayList<String>) params.get("columns");
+            for (int i = 0; i < columns.size(); i++) {
+                url += "columns[]=" + columns.get(i) + "&";
+            }
+            // Delete the columns from the param since it's added to the url query.
+            params.remove("columns");
+        }
+
+        if (params.containsKey("additional_columns")) {
+            List<String> additionalColumns = (ArrayList<String>) params.get("additional_columns");
+            for (int i = 0; i < additionalColumns.size(); i++) {
+                url += "additional_columns[]=" + additionalColumns.get(i) + "&";
+            }
+            // Delete the additional columns from the param since it's added to the url query.
+            params.remove("additional_columns");
+        }
+
         Map<String, Object> wrappedParams = new HashMap<String, Object>();
         wrappedParams.put("report", params);
         wrappedParams.put("start_date", params.get("start_date"));
         wrappedParams.put("end_date", params.get("end_date"));
 
-        String type = (String) params.get("type");
-        return request(RequestMethod.POST, reportURL(type), wrappedParams, Report.class, apiKey);
+        return request(RequestMethod.POST, url, wrappedParams, Report.class, apiKey);
     }
 
     /**
--- a/src/test/java/com/easypost/ReportTest.java
+++ b/src/test/java/com/easypost/ReportTest.java
@@ -1,5 +1,7 @@
 package com.easypost;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.List;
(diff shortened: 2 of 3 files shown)


src/main/java/com/easypost/model/Report.java
```diff
@@ -44,8 +48,33 @@ public static Report create(final Map<String, Object> params, final String apiKe
         wrappedParams.put("start_date", params.get("start_date"));
         wrappedParams.put("end_date", params.get("end_date"));
 
+        List<String> columns = new ArrayList<>();
+        List<String> additionalColumns = new ArrayList<>();
+
+
```
Justintime50, line 54: Seldom is there a need for 2 line breaks, let's just do one.

src/main/java/com/easypost/model/Report.java
```diff
@@ -44,8 +48,33 @@ public static Report create(final Map<String, Object> params, final String apiKe
...
+        if (params.containsKey("additional_columns")) {
+            additionalColumns = (ArrayList<String>) params.get("additional_columns");
+        }
+
         String type = (String) params.get("type");
-        return request(RequestMethod.POST, reportURL(type), wrappedParams, Report.class, apiKey);
+        String url = reportURL(type) + "?";
+
+        for (int i = 0; i < columns.size(); i++) {
+            url += "column[]=" + columns.get(i) + "&";
```
Justintime50, line 67: Please only construct the url params inside of the above if block. Only if the columns and additional columns keys are present do we want to alter the URL

src/main/java/com/easypost/model/Report.java
```diff
@@ -44,8 +48,33 @@ public static Report create(final Map<String, Object> params, final String apiKe
...
+        }
+
+        for (int i = 0; i < additionalColumns.size(); i++) {
+            url += "additional_column[]=" + additionalColumns.get(i) + "&";
```
Justintime50, line 71: Please only construct the url params inside of the above if block. Only if the columns and additional columns keys are present do we want to alter the URL

src/main/java/com/easypost/model/Report.java
```diff
@@ -44,8 +48,33 @@ public static Report create(final Map<String, Object> params, final String apiKe
...
+        }
+
+        Report response = request(RequestMethod.POST, url, wrappedParams, Report.class, apiKey);
+        response.setColumns(columns);
+        response.setAdditionalColumns(additionalColumns);
+        return response;
```
Justintime50, line 77: Nit: Add a line break before a return statement for easier readability

src/test/java/com/easypost/ReportTest.java
```diff
@@ -64,13 +66,36 @@ public static void setup() throws EasyPostException {
         globalTrackerReport = Report.create(trackerParams);
     }
 
+    @Test
+    public void testCreateReportWithAdditionalColumns() throws EasyPostException {
```
Justintime50, line 70: This is fine, please also add a test that only passes the `columns` param.

jchen293: added

nwithan8 (approved):

nwithan8 (dismissed):

src/test/java/com/easypost/ReportTest.java
```diff
@@ -64,16 +66,59 @@ public static void setup() throws EasyPostException {
...
+
+        reportWithAdditionalColumnsParam.put("start_date", Fixture.reportStartDate());
+        reportWithAdditionalColumnsParam.put("end_date", Fixture.reportEndDate());
+        reportWithAdditionalColumnsParam.put("type", "shipment");
+        reportWithAdditionalColumnsParam.put("additional_columns", additionalColumns);
+
+        Report reportWithAdditionalColumns = Report.create(reportWithAdditionalColumnsParam);
+
+        assertTrue(reportWithAdditionalColumns instanceof Report);
+        assertTrue(reportWithAdditionalColumns.getId().startsWith("shprep_"));
```
nwithan8, line 88: We don't need this ID test

src/test/java/com/easypost/ReportTest.java
```diff
@@ -64,16 +66,59 @@ public static void setup() throws EasyPostException {
...
+
+        reportWithAdditionalColumnsParam.put("start_date", Fixture.reportStartDate());
+        reportWithAdditionalColumnsParam.put("end_date", Fixture.reportEndDate());
+        reportWithAdditionalColumnsParam.put("type", "shipment");
+        reportWithAdditionalColumnsParam.put("columns", columns);
+
+        Report reportWithColumns = Report.create(reportWithAdditionalColumnsParam);
+
+        assertTrue(reportWithColumns instanceof Report);
+        assertTrue(reportWithColumns.getId().startsWith("shprep_"));
```
nwithan8, line 110: We don't need this ID check

nwithan8 (changes requested):

src/test/java/com/easypost/ReportTest.java
jchen293: I believe from the discussion yesterday, we are going to have these two assertions, similar to other unit tests for `report`.

Justintime50: Not exactly, we don't need to assert the ID since all we are testing is if the report gets created correctly. We are testing the ID elsewhere. We should assert that what comes back is a report (indicating success with the params passed) and then leaving a comment similar to what we do with batches saying we can't test this more in-depth because the report gets queued for generation and most likely isn't available quick enough for us to retrieve.

jchen293: sounds good, this will be addressed

src/test/java/com/easypost/ReportTest.java
jchen293: addressed

Justintime50 (approved):

PR #109, opened 2022-03-23
move columns and additional_columns from url query to request body.

jchen293:
This PR moves the `columns` and `additional_columns` from URL query to request body because a long query URL can be rejected or silently truncated. 

E2E test:
Screen Shot 2022-03-23 at 4 56 22 PM

### DIFF
--- a/src/main/java/com/easypost/model/Report.java
+++ b/src/main/java/com/easypost/model/Report.java
@@ -8,7 +8,6 @@ import java.net.URLEncoder;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.List;
 import java.util.ArrayList;
 
 public final class Report extends EasyPostResource {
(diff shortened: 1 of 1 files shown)


src/main/java/com/easypost/model/Report.java
```diff
@@ -42,32 +41,21 @@ public static Report create(final Map<String, Object> params) throws EasyPostExc
...
         }
 
         if (params.containsKey("additional_columns")) {
-            List<String> additionalColumns = (ArrayList<String>) params.get("additional_columns");
-            for (int i = 0; i < additionalColumns.size(); i++) {
-                url += "additional_columns[]=" + additionalColumns.get(i) + "&";
-            }
-            // Delete the additional columns from the param since it's added to the url query.
-            params.remove("additional_columns");
+            wrappedParams.put("additional_columns", params.getOrDefault("additional_columns", new ArrayList<String>()));
```
nwithan8, line 55: We only step into this line if `additional_columns` exists as a key in `params`, so planning for a default seems redundant.

Instead, let's just do:
```java
wrappedParams.put("additional_columns", params.getOrDefault("additional_columns", new ArrayList<String>()));
```
below `end_date` on line 48, and get rid of this if statement. Same for `columns`

In fact, we don't even need to add pretty much all the params from `params` to `wrappedParams`. We could just remove `type` from params and then use the `params` object as is.

nwithan8 (changes requested):

nwithan8 (approved):

Justintime50 (approved):

PR #110, opened 2022-03-24
minor improvement on report creation.

jchen293:
This PR removes the `wrapperParam` to make the code cleaner, add an exception when the user input does not contain `type` for report creation.

### DIFF
--- a/src/main/java/com/easypost/model/Report.java
+++ b/src/main/java/com/easypost/model/Report.java
@@ -8,7 +8,6 @@ import java.net.URLEncoder;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.ArrayList;
 
 public final class Report extends EasyPostResource {
     private String id;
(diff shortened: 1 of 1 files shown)


src/main/java/com/easypost/model/Report.java
```diff
@@ -40,22 +39,16 @@ public static Report create(final Map<String, Object> params) throws EasyPostExc
...
      */
     public static Report create(final Map<String, Object> params, final String apiKey) throws EasyPostException {
-        String type = (String) params.get("type");
-
-        Map<String, Object> wrappedParams = new HashMap<String, Object>();
-        wrappedParams.put("report", params);
-        wrappedParams.put("start_date", params.get("start_date"));
-        wrappedParams.put("end_date", params.get("end_date"));
+        if (params.containsKey("type")) {
+            String type = (String) params.get("type");
```
Justintime50, line 43: Since this is not reused, we can simply pass in `params.get("type")` to the `reportURL` param below.

src/main/java/com/easypost/model/Report.java
```diff
@@ -40,22 +39,16 @@ public static Report create(final Map<String, Object> params) throws EasyPostExc
...
-            wrappedParams.put("columns", params.get("columns"));
-        }
+            HashMap<String, Object> paramsWithoutType = new HashMap<>(params);
+            paramsWithoutType.remove("type");
 
-        if (params.containsKey("additional_columns")) {
-            wrappedParams.put("additional_columns", params.get("additional_columns"));
+            return request(RequestMethod.POST, reportURL(type), paramsWithoutType, Report.class, apiKey);
+        } else {
+            throw new EasyPostException("Please input a type for the report creation.");
```
Justintime50, line 50:
```suggestion
            throw new EasyPostException("Report type is required.");
```

Justintime50 (changes requested):

Justintime50 (approved):

nwithan8 (approved):

PR #116, opened 2022-04-07
removes the dead message conditional check in address create and verify.

jchen293:
- Removes dead `message` conditional check in Address `create and verify`
- Updates a few address assertions to match internal API changes

### DIFF
--- a/src/test/java/com/easypost/AddressTest.java
+++ b/src/test/java/com/easypost/AddressTest.java
@@ -109,7 +109,7 @@ public class AddressTest {
 
         assertTrue(address instanceof Address);
         assertTrue(address.getId().startsWith("adr_"));
-        assertEquals("417 MONTGOMERY ST STE 500", address.getStreet1());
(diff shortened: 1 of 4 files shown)


src/test/java/com/easypost/AddressTest.java
```diff
@@ -109,7 +109,7 @@ public void testCreateVerify() throws EasyPostException {
 
         assertTrue(address instanceof Address);
         assertTrue(address.getId().startsWith("adr_"));
-        assertEquals("417 MONTGOMERY ST STE 500", address.getStreet1());
```
Justintime50, line 112: These should be `FL 5`, please correct the spelling of `streat` in the fixture itself and re-record cassettes.

jchen293: I have updated the fixture data. There is no VCR in Java CL yet, E2E tested.

Justintime50 (changes requested):

Justintime50 (approved):

jontsai (approved):

PR #120, opened 2022-04-13
remove additional code that add strict and verify_strict to url query.

jchen293:
# Pull Request Type

# Description

This PR removes code that adds `strict` and `verify_strict` to the URL query when the API method is `POST`. The `wrappedParams` is also removed since it does not do anything. 

# Testing

Since Java CL is the only library that does not have VCR yet, E2E is locally tested.
Screen Shot 2022-04-13 at 2 26 52 PM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Run CheckStyle checks
-        uses: nikitasavinov/checkstyle-action@0.5.0
+        uses: nikitasavinov/checkstyle-action@0.5.1
         with:
           level: error
           fail_on_error: true
--- a/src/main/java/com/easypost/model/Address.java
+++ b/src/main/java/com/easypost/model/Address.java
@@ -4,7 +4,6 @@ import com.easypost.exception.EasyPostException;
 import com.easypost.net.EasyPostResource;
 
 import java.util.HashMap;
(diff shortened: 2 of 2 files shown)


Justintime50 (changes requested):
Code looks fine. Please add a test for that uses the `verify` param, we appear to be missing that in this lib. Look at any of the others for an example and then I'll stamp.

jchen293:
> Code looks fine. Please add a test for that uses the `verify` param, we appear to be missing that in this lib. Look at any of the others for an example and then I'll stamp.

There is a test that uses `verify` param: https://github.com/EasyPost/easypost-java/blob/master/src/test/java/com/easypost/AddressTest.java#L123

The `verify` param is in the fixture.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,12 +1,3 @@
-# Pull Request Type
```
Justintime50, line 1: This isn't related, please do this separately.

Justintime50 (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
jchen293: ok

Justintime50 (approved):

nwithan8 (approved):
So much cleaner now 👍🏻

PR #121, opened 2022-04-14
remove additional code that add strict and verify_strict to url query.

jchen293:
# Pull Request Type

# Description

Add wrapperParam when creating an address and put `verification` in the top-level param instead of inside the `address` param. 

Minor change in the fixture file and adjust unit test.

# Testing

E2E tested it, see screenshot below.

Screen Shot 2022-04-14 at 2 44 43 PM

### DIFF
--- a/src/test/java/com/easypost/AddressTest.java
+++ b/src/test/java/com/easypost/AddressTest.java
@@ -51,9 +51,9 @@ public class AddressTest {
     @Test
     public void testCreateVerifyStrict() throws EasyPostException {
         Map<String, Object> addressData = Fixture.basicAddress();
-        List<String> verificationList = new ArrayList<>();
+        List<Boolean> verificationList = new ArrayList<>();
         
-        verificationList.add("true");
+        verificationList.add(true);
         addressData.put("verify_strict", verificationList);
 
         Address address = Address.create(addressData);
@@ -105,7 +105,13 @@ public class AddressTest {
      */
     @Test
     public void testCreateVerify() throws EasyPostException {
-        Address address = Address.create(Fixture.incorrectAddressToVerify());
+        List<Boolean> verificationList = new ArrayList<>();
+        verificationList.add(true);
+
+        Map<String, Object> addressData = Fixture.incorrectAddressToVerify();
+        addressData.put("verify", verificationList);
+
+        Address address = Address.create(addressData);
 
         assertTrue(address instanceof Address);
         assertTrue(address.getId().startsWith("adr_"));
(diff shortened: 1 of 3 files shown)


src/test/java/com/easypost/AddressTest.java
```diff
@@ -105,7 +105,13 @@ public void testAll() throws EasyPostException{
      */
     @Test
     public void testCreateVerify() throws EasyPostException {
-        Address address = Address.create(Fixture.incorrectAddressToVerify());
+        List<String> verificationList = new ArrayList<>();
+        verificationList.add("delivery");
```
Justintime50, line 109: Let's change this from `delivery` as a string to a boolean `true` so it matches our other libs. This way we can keep the assertions the same if ever the verification criteria changes in the future.

src/test/java/com/easypost/AddressTest.java
```diff
@@ -120,7 +126,13 @@ public void testCreateVerify() throws EasyPostException {
      */
     @Test
     public void testCreateAndVerify() throws EasyPostException {
-        Address address = Address.createAndVerify(Fixture.incorrectAddressToVerify());
+        List<String> verificationList = new ArrayList<>();
+        verificationList.add("delivery");
+
+        Map<String, Object> addressData = Fixture.incorrectAddressToVerify();
+        addressData.put("verify", verificationList);
```
Justintime50, line 133: For this method, we don't need to pass the `verify` param because this endpoint will automatically verify the address passed to it.

src/test/java/com/easypost/AddressTest.java
```diff
@@ -120,7 +126,13 @@ public void testCreateVerify() throws EasyPostException {
...
+
+        Address address = Address.createAndVerify(addressData);
```
Justintime50, line 135: You can pass the fixture directly to this function.

Justintime50 (changes requested):

src/test/java/com/easypost/AddressTest.java
jchen293: got it

Justintime50 (approved):

src/test/java/com/easypost/AddressTest.java
```diff
@@ -51,9 +51,9 @@ public void testCreate() throws EasyPostException {
     @Test
     public void testCreateVerifyStrict() throws EasyPostException {
         Map<String, Object> addressData = Fixture.basicAddress();
-        List<String> verificationList = new ArrayList<>();
+        List<Boolean> verificationList = new ArrayList<>();
         
-        verificationList.add("true");
+        verificationList.add(true);
```
nwithan8, line 56: Thank you!

nwithan8 (approved):

PR #122, opened 2022-04-14
Verification

jchen293:
# Pull Request Type

# Description

Add missing classes: `verification`, `verifications`, and `verificationDetails`
Add extra assertion for `createAndVerify` Address.
Correct the type of `Verifications` in Address class attribute, according to our [API Doc](https://www.easypost.com/docs/api#verification-object). 

# Testing

E2E tested locally

Screen Shot 2022-04-14 at 6 59 45 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Add support for `columns` and `additional_columns` on Report creation
 * Add `declaraction` attribute in CustomsInfo class
 * Remove the dead `message` conditional check in `Address.createAndVerify()`
+* Add `AddressVerifications` class
 
 ## v5.2.0 (2022-02-25)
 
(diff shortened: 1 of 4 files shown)


Justintime50 (approved):

Justintime50 (changes requested):
It appears we already have a bunch of this already, it's called `AddressVerification`, there's probably no need to change it here.

Also, CHANGELOG entries are supposed to be in present tense `Adds` instead of `Added`, etc. Please put them back or correct them to present tense. Most of the changes here may not be necessary however.

Justintime50 (approved):
Just to make sure I understand this, we are adding the missing bridge here from an `Address` object down to it's `Verification` object by adding the missing `Verifications` object?

So it is going from `Address -> Verification` which is incorrect to `Address -> Verifications -> Verification` which now makes it accessible?

jchen293:
> 

That's correct, so the user can access `zip4` and `delivery` from `Verification`, according to the API doc.

Screen Shot 2022-04-15 at 5 24 57 PM

nwithan8 (approved):
LGTM, mirrors what we have in [Go](https://github.com/EasyPost/easypost-go/blob/master/address.go) (which is what I'm going to start considering our reference for type-strict languages)

PR #128, opened 2022-05-10
Add explicit Smartrate class

nwithan8:
# Description

Previously, we didn't have an explicit Smartrate class, and were instead over-using the Rate class to store smartrate-only properties such as TimeInTransit.

This PR:
- Introduces a `Smartrate` class (mostly an extension of `Rate`)
- Improves the JSON deserializer to handle smartrate objects
- Replaces all smartrate-related functions (i.e. in `Shipment`) that previously returned `Rate` objects, now return `Smartrate` objects (potential breaking change only if end user was checking instance type)

# Testing

- Re-ran related unit tests, all pass
- Traced through code in debug mode, everything looks good

# Pull Request Type


nwithan8:
NOTE: The `Smartrate` class, while it adopts the setter/getter functions for all the standard `Rate` properties via its extension of the `Rate` class, does not have a setter (only a getter) for its `timeInTransit` property.

This is because a setter is not needed. JSON deserialization via Gson does not use setters, so the only use for a setter would be to allow the end-user a way to modify the `timeInTransit` property for a `Smartrate` object, which they likely would never need to do, nor should we allow them to do it.

This is part of a larger discussion about how many of our setters are actually warranted in our code, what properties do we want to give end-users the ability to alter, and which should be read-only.

src/main/java/com/easypost/model/Smartrate.java
```diff
@@ -0,0 +1,14 @@
+package com.easypost.model;
+
+public class Smartrate extends Rate {
+    private TimeInTransit timeInTransit;
+
+    /**
+     * Get the time in transit for this rate.
```
jchen293, line 7:
```suggestion
     * Get the time in transit for this smart rate.
```

src/main/java/com/easypost/model/Shipment.java
```diff
@@ -789,32 +789,32 @@ public Shipment newRates(final Map<String, Object> params, final String apiKey)
     /**
      * Get smart rates for this Shipment.
```
Justintime50, line 790: These should all be one word please

```suggestion
     * Get smartrates for this Shipment.
```

Justintime50 (commented):
Looks good, let's just correct the docstrings and combine the word `smartrate`

src/main/java/com/easypost/model/SmartrateCollection.java
```diff
@@ -4,39 +4,49 @@
 import java.util.List;
 
 public final class SmartrateCollection {
-    private final List<Rate> rates;
+
```
jchen293, line 7: no need a new line.

jchen293 (commented):
Can you also update the CHANGELOG as well?

jchen293 (approved):

Justintime50 (approved):

PR #129, opened 2022-05-10
clean up IDE build warnings and other minor changes.

jchen293:
# Description

Clean up IDE warnings during the build
Change the setter to be public access modifier

# Testing

locally built without any warning
Screen Shot 2022-05-11 at 11 49 00 AM

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,274 +13,274 @@
 
 ## v5.2.0 (2022-02-25)
 
--   Adds a getter and setter for `eel_pfc` on the `CustomsInfo` object
--   Adds a getter and setter for `id` on the `Brand` object
--   Fixes a bug that didn't allow Refunds to be created
--   Adds a complete test suite and upgrade JUnit4 to JUnit5
--   Fixes the `AddressCollection` to return a list of `Address` objects instead of `Batch` objects
+- Adds a getter and setter for `eel_pfc` on the `CustomsInfo` object
+- Adds a getter and setter for `id` on the `Brand` object
+- Fixes a bug that didn't allow Refunds to be created
+- Adds a complete test suite and upgrade JUnit4 to JUnit5
+- Fixes the `AddressCollection` to return a list of `Address` objects instead of `Batch` objects
 
 ## v5.1.0 (2022-02-09)
 
--   Adds support to one-call buy a Shipment by passing a `service` and `carrierAccounts` key
--   Adds support to retrieve carrier types via `CarrierType.all()`
--   Adds support to one-call buy an Order by passing a `service` and `carrierAccounts` key
--   Adds support to update the user's brand via `User.updateBrand()`
--   Allow user to override default API timeouts via getters and setters
--   Bump all plugins and remove dead `oss-parent` plugin
--   Various other small improvements and bug fixes
+- Adds support to one-call buy a Shipment by passing a `service` and `carrierAccounts` key
+- Adds support to retrieve carrier types via `CarrierType.all()`
+- Adds support to one-call buy an Order by passing a `service` and `carrierAccounts` key
+- Adds support to update the user's brand via `User.updateBrand()`
+- Allow user to override default API timeouts via getters and setters
+- Bump all plugins and remove dead `oss-parent` plugin
+- Various other small improvements and bug fixes
 
 ## v5.0.0 (2022-01-14)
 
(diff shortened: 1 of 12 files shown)


src/main/java/com/easypost/model/PostageLabel.java
```diff
@@ -39,7 +39,7 @@ public String getId() {
      *
      * @param id ID of this PostageLabel.
      */
-    private void setId(final String id) {
+    public void setId(final String id) {
```
nwithan8, line 42: Why were these all private to begin with? Does that mean users haven't been able to retrieve any `PostageLabel` properties up until now?

jchen293: It's `setter` only and `getter` is still public in this class previously. `PostageLabel` is the only class that has private `setters` and this needs to be public instead, though most users will probably not use `setter` functions.

nwithan8: Thanks for the clarification. Yeah, we're going to explore what properties need to have `setter` functions in the future, but for now, making them public and consistent is good.

Justintime50 (dismissed):
Thank you for cleaning up all the build errors! 😃  I can confirm they are gone.

There are still a few IDE errors though. Try opening this in VS Code and you'll see, let's clean these up too please:

Screen Shot 2022-05-11 at 9 58 47 AM

src/main/java-templates/com/easypost/package-info.java
```diff
@@ -1,9 +0,0 @@
-/**
```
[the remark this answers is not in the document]
nwithan8: No, the original error was because this existed, but as part of the build process, it autogenerates this file as well, throwing the warning because there were then two of them.

CHANGELOG.md
```diff
@@ -13,274 +13,274 @@
 
 ## v5.2.0 (2022-02-25)
 
--   Adds a getter and setter for `eel_pfc` on the `CustomsInfo` object
```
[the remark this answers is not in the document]
jchen293: It should be one space instead of three, otherwise, there will be a warning in the IDE.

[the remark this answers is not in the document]
Justintime50: Ah it appears I have an aggressive formatting plugin for markdown and this was a bug: https://github.com/prettier/prettier/issues/5019, let's keep this fix and I'll fix my own formatter on my end.

nwithan8 (dismissed):

Justintime50 (changes requested):
Please add `.vscode` to the `.gitignore` and I'll stamp! Needed to remove IDE warnings for VS code when setting some settings to remove warnings due to imports on a rebuild

src/main/java/com/easypost/app/Readme.java
```diff
@@ -67,6 +67,7 @@ public static void main(final String[] args) {
 
 
             try {
+                @SuppressWarnings("unused")
                 Address verified = toAddress.verify();
```
Justintime50, line 71:
```suggestion
                toAddress.verify();
```

src/test/java/com/easypost/PickupTest.java
```diff
@@ -74,7 +74,7 @@ public void testRetrieve() throws EasyPostException {
     @Test
     @Disabled // This test is ignored for now because it relies on the pickup min and max date
     public void testBuy() throws EasyPostException {
-        // TODO We will revisit this once we have VCR.
+        // todo: We will revisit this once we have VCR.
```
Justintime50, line 77: Let's just remove this. We'll need to adjust this when we do VCR anyway.

src/test/java/com/easypost/PickupTest.java
```diff
@@ -96,7 +96,7 @@ public void testBuy() throws EasyPostException {
     @Test
     @Disabled // This test is ignored for now because it relies on the pickup min and max date
     public void testCancel() throws EasyPostException {
-        // TODO We will revisit this once we have VCR.
+        // todo: We will revisit this once we have VCR.
```
Justintime50, line 99: Remove this line

Justintime50 (changes requested):

Justintime50 (approved):
Can verify no build warnings/errors and no IDE warnings/errors.

Fantastic work on this!

nwithan8 (approved):

PR #130, opened 2022-05-12
End shipper

jchen293:
# Description

Adds the new EndShipper beta class with create, retrieve, all, and update functions. This is the fully-qualified address object that the USPS will start requiring shortly.

# Testing

Write new unit tests

# Pull Request Type


src/test/java/com/easypost/Fixture.java
```diff
@@ -343,4 +344,26 @@ public static Map<String, Object> basicOrder() {
 
         return order;
     }
+
+    /**
+     * Basic end shipper address that is reusable in all test coverage.
```
Justintime50, line 349: We should be exact when referencing EasyPost objects.

```suggestion
     * Basic EndShipper address that is reusable in all test coverage.
```

src/test/java/com/easypost/Fixture.java
```diff
@@ -343,4 +344,26 @@ public static Map<String, Object> basicOrder() {
...
+     *
+     * @return A map that has the end shipper address info filled.
```
Justintime50, line 351:
```suggestion
     * @return A map that has the EndShipper address info filled.
```

src/test/java/com/easypost/Fixture.java
```diff
@@ -343,4 +344,26 @@ public static Map<String, Object> basicOrder() {
...
+     */
+    public static Map<String, Object> basicEndShipperAddress() {
```
Justintime50, line 353: Just call it `endShipperAddress` because all fields are always required meaning there won't ever be multiple kinds based on the filled in data.

```suggestion
    public static Map<String, Object> endShipperAddress() {
```

Justintime50 (changes requested):
Great working! 2 tiny name updates and this is ready 🙂

Justintime50 (dismissed):

src/test/java/com/easypost/beta/EndShipperTest.java
```diff
@@ -0,0 +1,100 @@
+package com.easypost.beta;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.List;
+
+import com.easypost.EasyPost;
+import com.easypost.Fixture;
+import com.easypost.exception.EasyPostException;
+import com.easypost.model.beta.EndShipper;
```
Justintime50, line 10: One quick Q: Is this standard in Java to place modules inside this `model` directory? If so, can you provide some documentation on the subject here for posterity?

jchen293: It's not a standard in Java. I created the `beta` folder in the `model` directory because that's where all the classes are hosted, also being consistent with other libs Python and Ruby where we put the `beta` folder in the same directory.

src/main/java/com/easypost/model/beta/EndShipper.java
```diff
@@ -0,0 +1,314 @@
...
+import com.easypost.EasyPost;
+import com.easypost.exception.EasyPostException;
+import com.easypost.net.EasyPostResource;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public final class EndShipper extends EasyPostResource {
```
nwithan8, line 12: I thought we were going to extend the Address class, and then simply override the address methods, since ultimately an `EndShipper` is really just an `Address`?

src/test/java/com/easypost/Fixture.java
```diff
@@ -5,13 +5,9 @@
 import java.util.List;
 import java.util.Map;
 
-final class Fixture {
+public abstract class Fixture {
```
nwithan8, line 8: Why did this get switched to `abstract`? Is it being extended/re-implemented somewhere else, and if so, why?

nwithan8 (changes requested):

src/test/java/com/easypost/Fixture.java
jchen293: This class is being used in the unit tests. Changed it to `abstract` to prevent instantiation.

src/main/java/com/easypost/model/beta/EndShipper.java
jchen293: Yes, we can do that. But I don't want users to access other functions from the `Address` class such as `create and verify()`

nwithan8: You can override those functions and simply throw a NotImplemented exception

src/test/java/com/easypost/Fixture.java
nwithan8: Fair enough

src/main/java/com/easypost/model/beta/EndShipper.java
jchen293: Personally, I don't like overriding every method(except getters and setters) from the parent class `Address`.

Justintime50: Per our convo, let's decouple the EndShipper object from the Address object (do not extend). This has been corrected in Ruby.

src/main/java/com/easypost/model/Address.java
```diff
@@ -1,25 +1,14 @@
...
+public final class Address extends BaseAddress {
 
-    private String id;
     private String mode;
-    private String name;
-    private String company;
-    private String street1;
-    private String street2;
-    private String zip;
     private String city;
```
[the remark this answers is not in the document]
jchen293: fixed

src/main/java/com/easypost/model/BaseAddress.java
```diff
@@ -0,0 +1,215 @@
...
+
+    /**
+     * Set city of address.
+     *
+     * @param city address city
+     */
+    public void setCity(final String city) {
+        this.city = city;
+    }
+}
```
Justintime50, line 215: We should order this where it makes sense in the hierarchy of the rest of them (after street and before state)

jchen293: fixed

src/main/java/com/easypost/model/BaseAddress.java
```diff
@@ -0,0 +1,215 @@
...
+public class BaseAddress extends EasyPostResource {
+    private String id;
+    private String name;
+    private String company;
+    private String street1;
+    private String street2;
+    private String zip;
+    private String city;
+    private String state;
+    private String country;
```
Justintime50, line 14: We should list these in the order of a real address:

street1
street2
city
state
zip
country

This goes for the getters/setters too.

Justintime50 (changes requested):

src/main/java/com/easypost/model/BaseAddress.java
jchen293: fixed

nwithan8 (approved):

PR #132, opened 2022-05-19
Referral

jchen293:
# Description

Adds beta Referral class for White Label API with these new functions: `create()`, `updateEmail()`, `all()`, and `addCreditCard()`
Adds `BaseUser` class that `Referral` and `User` classes can extend from
Adds `CreditCard` class for `addCreditCard()` response type

# Testing

E2E test
Screen Shot 2022-05-19 at 12 43 32 PM

# Pull Request Type


nwithan8 (approved):

src/main/java/com/easypost/model/BaseUser.java
```diff
@@ -0,0 +1,179 @@
+package com.easypost.model;
+
+import com.easypost.net.EasyPostResource;
+
+import java.util.List;
+
+public class BaseUser extends EasyPostResource {
```
Justintime50, line 7: BaseUser should contain ALL the properties of User/Referral because they are the exact same option, they just have different methods. So either we need to have `BaseUser` contain all the properties (User/Referral would then inherit without adding any unique properties and just have methods) or we need to remove BaseUser completely and simply have Referral extend User.

Justintime50 (changes requested):

src/main/java/com/easypost/model/BaseUser.java
jchen293: `Referral` class has extra properties that the `User` class doesn't have, according to the internal doc.

Justintime50: Such as? They are identical objects that use different endpoints. Not every property of a User is documented on our public facing docs (which would explain the discrepancy between the User docs and the Referral docs).

This needs to be adjusted.

jchen293: Ooo, I think you're right. Looks like the `User` class only has partial properties.

src/main/java/com/easypost/model/beta/Referral.java
```diff
@@ -0,0 +1,303 @@
...
+        String easypostStripeApiKey = retrieveEasypostStripeApiKey();
+        String stripeToken;
+
+        try {
+            stripeToken = createStripeToken(number, expirationMonth, expirationYear, cvc, easypostStripeApiKey);
+        } catch (Exception e) {
+            throw new Exception("Could not send card details to Stripe, please try again later", e);
+        }
+
+        return createEasypostCreditCard(referralApiKey, stripeToken, priority.toString());
```
Justintime50, line 155: Let's go ahead and lowercase the `priority` here since the enum is all caps - just in case.

src/main/java/com/easypost/model/beta/Referral.java
```diff
@@ -0,0 +1,303 @@
...
+
+    /**
+     * Create Encoded URL from a Map.
+     *
+     * @param params Map of parameters to be encoded.
+     * @param parentKey Parent key in the encoded URL.
+     * @return Encoded URL for Stripe API call.
+     * @throws Exception
+     */
+    private static String getEncodedURL(Map<String, String> params, String parentKey)
```
Justintime50, line 262: Let's move this to a `Util` file because it's by nature a utility for encoding URLs and can be reused in the future if necessary. Since we are adding it brand new, let's put it in the right place to begin with.

src/test/java/com/easypost/Fixture.java
```diff
@@ -366,4 +366,38 @@ public static Map<String, Object> endShipperAddress() {
...
+     * @return A map that has the credit card info filled.
+     */
+    public static Map<String, String> creditCardDetails() {
+        // The credit card details below are for a valid proxy card usable
+        // for tests only and cannot be used for real transactions.
+        // DO NOT alter these details with real credit card information.
+        Map<String, String> creditCardDetails = new HashMap<String, String>();
+
+        creditCardDetails.put("number", "4536410136126170");
+        creditCardDetails.put("expiration_month", "5");
```
Justintime50, line 397: This should be `05`, Stripe expects a two digit entry.

Justintime50 (changes requested):

Justintime50 (approved):

nwithan8 (approved):

PR #136, opened 2022-05-27
adds user billing functions.

jchen293:
# Description

Adds `PaymentMethod.all()`, `CreditCard.fund()`, and `CreditCard.delete()` functions.
Adds `BaseCreditCard` class that other classes can extend from

# Testing

Added unit test for retrieving all payment methods
Can't test `CreditCard.fund()`, and `CreditCard.delete()` due to not having a valid CC to use

# Pull Request Type


src/main/java/com/easypost/model/CreditCard.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Fund a credit card.
+     *
+     * @param primaryOrSecondary primary or secondary payment method.
+     * @param amount amount to fund.
+     * @return CreditCardFund object.
+     * @throws EasyPostException when the request fails.
+     */
+    public static CreditCardFund fund(Priority primaryOrSecondary, int amount) throws EasyPostException {
```
Justintime50, line 23: Let's swap the param order. Amount first, primaryOrSecondary second (to match Python)

src/main/java/com/easypost/model/CreditCard.java
```diff
@@ -0,0 +1,78 @@
...
+     * @return CreditCardFund object.
+     * @throws EasyPostException when the request fails.
+     */
+    public static CreditCardFund fund(Priority primaryOrSecondary, int amount, String apiKey) throws EasyPostException {
+        PaymentMethod paymentMethods = PaymentMethod.all();
+        String cardID = null;
+
+        if (primaryOrSecondary == Priority.PRIMARY) {
+            cardID = paymentMethods.getPrimaryPaymentMethod().getId();
+        } else {
```
Justintime50, line 42: Let's be more explicit here and do an else if and check if it matches `SECONDARY`

src/main/java/com/easypost/model/CreditCardFund.java
```diff
@@ -0,0 +1,42 @@
+package com.easypost.model;
```
Justintime50, line 1: Why is this an object? I don't believe this is an actual EasyPost object.

src/main/java/com/easypost/model/PaymentMethod.java
```diff
@@ -0,0 +1,106 @@
...
+import com.easypost.exception.EasyPostException;
+import com.easypost.net.EasyPostResource;
+
+public class PaymentMethod extends EasyPostResource {
+    private String id;
+    private String object;
+    private PrimaryPaymentMethod primaryPaymentMethod;
+    private SecondaryPaymentMethod secondaryPaymentMethod;
+
+    /**
```
Justintime50, line 13: Functions need to come after the properties.

Justintime50 (changes requested):

src/main/java/com/easypost/model/CreditCardFund.java
jchen293: This is the return response after the user funds a credit card. We need this class for GSON serialization.

src/main/java/com/easypost/model/CreditCard.java
nwithan8: Or even better, a switch statement

src/main/java/com/easypost/model/PaymentMethod.java
nwithan8: I would have thought the linter would have caught this.

Justintime50: As would I. These are actually out of order all throughout the lib. I've created an internal task to circle back to cleaning these up.

src/main/java/com/easypost/model/CreditCard.java
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * Fund a credit card.
+     *
+     * @param amount amount to fund.
+     * @param primaryOrSecondary primary or secondary payment method.
+     * @param apiKey API key to use in request (overrides default API key).
+     * @return CreditCardFund object.
+     * @throws EasyPostException when the request fails.
+     */
+    public static CreditCardFund fund(int amount, Priority primaryOrSecondary, String apiKey) throws EasyPostException {
```
Justintime50, line 36: `amount` should be a string based on the docs. This will allow for decimals as well. Our API will handle it all behind the scenes.

Justintime50 (changes requested):

src/main/java/com/easypost/model/CreditCard.java
jchen293: fixed.

Justintime50 (approved):

src/main/java/com/easypost/model/PrimaryPaymentMethod.java
```diff
@@ -0,0 +1,4 @@
+package com.easypost.model;
+
+public class PrimaryPaymentMethod extends BaseCreditCard {
```
nwithan8, line 3: Nit: If there's nothing new added here, we don't really need explicit classes for `PrimaryPaymentMethod` and `SecondaryPaymentMethod`, we can just use `BaseCreditCard`.
Won't block, because we can keep them in case properties diverge in the future.

src/main/java/com/easypost/model/beta/CreditCard.java
```diff
@@ -1,137 +1,6 @@
...
-
-    /**
-     * Set the brand of this CreditCard object.
-     *
-     * @param brand the brand of this CreditCard.
-     */
-    public void setBrand(String brand) {
-        this.brand = brand;
-    }
+public class CreditCard extends BaseCreditCard {
```
nwithan8, line 5: Nit: same here, `CreditCard` is no different from `BaseCreditCard`, so it's not really needed. Won't block.

nwithan8 (approved):

PR #141, opened 2022-06-03
add billingType in Rate and CarrierAccount classes.

jchen293:
# Description

Add `billingType` in CarrierAccount and Rate classes

# Testing

# Pull Request Type

### DIFF
--- a/src/main/java/com/easypost/model/CreditCard.java
+++ b/src/main/java/com/easypost/model/CreditCard.java
@@ -13,7 +13,7 @@ public class CreditCard extends BaseCreditCard {
     }
 
     /**
-     * Fund a credit card.
+     * Fund your EasyPost wallet by charging your primary or secondary card on file.
      *
      * @param amount amount to fund.
      * @param primaryOrSecondary primary or secondary payment method.
@@ -25,7 +25,7 @@ public class CreditCard extends BaseCreditCard {
     }
 
     /**
-     * Fund a credit card.
+     * Fund your EasyPost wallet by charging your primary or secondary card on file.
      *
      * @param amount amount to fund.
      * @param primaryOrSecondary primary or secondary payment method.
@@ -61,7 +61,7 @@ public class CreditCard extends BaseCreditCard {
     }
 
     /**
-     * Delete a credit card.
+     * Delete a credit card by ID.
      *
      * @param creditCardId ID of credit card to delete.
      * @throws EasyPostException when the request fails.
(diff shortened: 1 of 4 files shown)


nwithan8 (approved):

src/main/java/com/easypost/model/Rate.java
```diff
@@ -334,4 +335,23 @@ public String getCarrierAccountId() {
     public void setCarrierAccountId(final String carrierAccountId) {
         this.carrierAccountId = carrierAccountId;
     }
+
+
```
Justintime50, line 339: Single space instead of double

src/main/java/com/easypost/model/CreditCard.java
```diff
@@ -25,7 +25,7 @@ public static CreditCardFund fund(String amount, Priority primaryOrSecondary) th
     }
 
     /**
-     * Fund a credit card.
+     * Fund your EasyPost wallet by charging your primary or secondary card on file.
```
Justintime50, line 28: Would have preferred these cleanup items in a separate diff since they don’t have anything to do with the title or summary of the changes in question.

src/main/java/com/easypost/model/Rate.java
```diff
@@ -20,6 +20,7 @@ public class Rate extends EasyPostResource {
     private Number estDeliveryDays;
     private String shipmentId;
     private String carrierAccountId;
+    private String billingType;
```
Justintime50, line 23: Does the smartrate object need this too?

src/main/java/com/easypost/model/CreditCard.java
jchen293: I will update the summary since this is a very tiny cleanup.

src/main/java/com/easypost/model/Rate.java
ldewald: Can a formatter be set up to avoid stuff like this in the future? I'm a fan of prettier and it looks like they have a plugin for Java

src/main/java/com/easypost/model/Rate.java
nwithan8: Doesn't look like it. There's no `billing_type` key in the JSON response for the "smartrate" endpoint (last entry in this cassette: https://github.com/EasyPost/easypost-java/blob/master/cassettes/shipment/lowest_smartrate.json)

jchen293: I also verify in the cURL command, no `billing_type` was return in the JSON response

nwithan8: This needs to be removed then.

nwithan8 (approved):

ldewald (approved):

PR #144, opened 2022-06-16
Implement getter/setter styling

nwithan8:
# Description

This PR adds changes to the `.editorconfig` file (described below). 
NOTE: Styling via `.editorconfig` is separate from CheckStyle, meaning that these styling rules can't necessarily be enforced by CI. But, many IDEs can import an `.editorconfig` file and use the styling rules during code rearrangement. Some IDEs can even run code rearrangement as a pre-commit step. Basically, please do your due diligence to adhere to these rules, and leverage your IDE's ability to do this automagically for you.

- Updated .editorconfig to enforce element order in .java files
- Getter/setters are now between properties and constructors, alphabetized
- Enums are now at the top of class files

# Testing

No code changes, just re-ordered elements.
All tests pass as expected.

# Pull Request Type


src/main/java-templates/com/easypost/EasyPost.java
```diff
@@ -12,7 +12,7 @@ public abstract class EasyPost {
 
     /**
      * Set a VCR to be used for all HTTP requests.
-     *
+     * <p>
```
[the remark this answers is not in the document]
Justintime50: It appears that this is the spec and should be fine:

Screen Shot 2022-06-16 at 2 21 49 PM

src/main/java/com/easypost/model/CarrierAccount.java
```diff
@@ -241,6 +223,24 @@ public CarrierAccount update(final Map<String, Object> params, final String apiK
         return this;
     }
 
+    /**
```
Justintime50, line 226: The ID getter/setter isn't in the right place? With the rules changes, shouldn't Checkstyle/your IDE complain about this?

src/main/java/com/easypost/model/CreditCard.java
```diff
@@ -74,11 +70,16 @@ public static void delete(String creditCardId) throws EasyPostException {
...
      * @throws EasyPostException when the request fails.
      */
     public static void delete(String creditCardId, String apiKey) throws EasyPostException {
-        request(RequestMethod.DELETE, String.format("%s/%s/%s", EasyPost.API_BASE, "credit_cards",
-            creditCardId), null, CreditCard.class, apiKey);
+        request(RequestMethod.DELETE, String.format("%s/%s/%s", EasyPost.API_BASE, "credit_cards", creditCardId), null,
+                CreditCard.class, apiKey);
+    }
+
+    public enum Priority {
```
Justintime50, line 81: I'd expect enums to be at a top of a file, not the very bottom.

src/main/java/com/easypost/model/Event.java
```diff
@@ -77,39 +31,39 @@ public void setId(final String id) {
...
      */
-    public void setDescription(final String description) {
-        this.description = description;
+    public void setMode(final String mode) {
+        this.mode = mode;
     }
 
     /**
-     * Get the mode of this Event.
+     * Get the description of this Event.
```
Justintime50, line 52: The getters/setters don't match the same order of the properties.

What do you suggest here? We can make the getters/setters alphabetical (will mismatch the property order) or ensure the getter/setter order matches the property ordering for consistency.

Justintime50 (dismissed):

src/main/java/com/easypost/model/Event.java
nwithan8: I was wondering that myself. I don't think there's a way to match the order of properties and the getter/setters (I was hoping it did, but seems it didn't).

The former, alphabetizing them, is enforceable by .editorconfig

src/main/java/com/easypost/model/CreditCard.java
nwithan8: Sounds good, will bump.

src/main/java/com/easypost/model/CarrierAccount.java
nwithan8: Odd, I'll check it out.

nwithan8: I'm not sure. There's getter/setters for `id` in the base `EasyPostResource` class, but this seems to be the only one that gets tripped up by the override method.

NOTE FOR FUTURE: We shouldn't be overriding the getter/setters for `id` for no real reason...

src/main/java/com/easypost/model/Event.java
nwithan8: Alphabetical it is for now. We can revisit this when we clean up inherited properties in the future.

Justintime50 (approved):

jchen293 (approved):

PR #149, opened 2022-06-21
Rename non-getter `getX` methods

nwithan8:
# Description

To avoid confusion between proper getter/setter functions and other functions that merely start with the word `get`, all non-getter `getX` functions have been renamed to simply `X` (i.e. `getLowestSmartrate()` -> `lowestSmartrate()`).

This will avoid unnecessary exception throwing and catching during object merges, which utilizes getter/setter pairs and gets confused when there is no `setX` for a given `getX` function.

# Testing

All tests pass as expected.

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,7 @@
 - Some `getX` functions have been deprecated and renamed to simply `X` to avoid confusion with proper getter/setter
   functions.
     - `getSmartrates()` -> `smartrates()`
-    - `getLowestSmartrate()` -> `lowestSmartrate()`
-    - `getSmartRateAccuracy()` -> `smartrateAccuracy()`
+    - `getLowestSmartRate()` -> `findLowestSmartrate()`
 
 ## v5.5.0 (2022-06-21)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # CHANGELOG
 
+## Next Release
+
+- Some `getX` functions have been deprecated and renamed to simply `X` to avoid confusion with proper getter/setter
+  functions.
+    - `getSmartrates()` -> `smartrates()`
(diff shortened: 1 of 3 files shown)


src/main/java/com/easypost/model/CarrierAccount.java
```diff
@@ -111,6 +147,60 @@ public void setReference(final String reference) {
...
+    public void setType(String type) {
+        this.type = type;
+    }
+
+    /**
+     * Get whether the carrier account is a clone.
+     *
+     * @return True if carrier account is a clone, false otherwise.
+     */
+    public boolean isClone() {
```
Justintime50, line 191: Shouldn't this be `get` in a perfect world?

Justintime50 (changes requested):
The goal is not to make a breaking change here but to keep the original functions present but deprecate them and point users to the new (correctly named) items. Then when we have a major release, we can remove the deprecated items and the odd workaround.

Justintime50 (dismissed):
There's a ton of changes not related at all to the smartrate deprecation. Please remove or pull unrelated refactoring out into a separate PR. This PR should really only touch 2 files.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # CHANGELOG
 
+## Next Release
+
+- Some `getX` functions have been deprecated and renamed to simply `X` to avoid confusion with proper getter/setter
+  functions.
+    - `getSmartrates()` -> `smartrates()`
+    - `getLowestSmartrate()` -> `lowestSmartrate()`
+    - `getSmartRateAccuracy()` -> `smartrateAccuracy()`
```
Justintime50, line 9: This line needs to be removed since it's a getter and not getting removed.

Justintime50 (approved):

jchen293 (approved):

PR #152, opened 2022-06-29
Migrate to using enums for Smartrate delivery accuracy choices

nwithan8:
# Description

- New SmartrateAccuracy enum (similar to [C# library](https://github.com/EasyPost/easypost-csharp/blob/master/EasyPost/Smartrate.cs))
- New functions to use SmartrateAccuracy when dealing with smartrates
- Deprecate old smartrate functions using strings

# Testing

- Unit tests updated to use new functions
- No new unit tests
- No new cassettes needed
- No new recordings needed

# Pull Request Type

### DIFF
--- a/src/main/java/com/easypost/model/Shipment.java
+++ b/src/main/java/com/easypost/model/Shipment.java
@@ -1086,27 +1086,11 @@ public final class Shipment extends EasyPostResource {
      * @param deliveryAccuracy Delivery days accuracy restriction to use when filtering.
      * @return lowest Smartrate object
      * @throws EasyPostException when the request fails.
-     * @deprecated Use {@link #findLowestSmartrate(List, int, String)} instead.
+     * @deprecated Use {@link #findLowestSmartrate(List, int, SmartrateAccuracy)} instead.
      */
     @Deprecated
     public static Smartrate getLowestSmartRate(final List<Smartrate> smartrates, int deliveryDay,
                                                String deliveryAccuracy) throws EasyPostException {
-        return findLowestSmartrate(smartrates, deliveryDay, deliveryAccuracy);
-    }
-
-    /**
-     * Find the lowest Smartrate from a list of Smartrates.
-     *
-     * @param smartrates       List of Smartrates to filter from.
-     * @param deliveryDay      Delivery days restriction to use when filtering.
-     * @param deliveryAccuracy Delivery days accuracy restriction to use when filtering.
-     * @return lowest Smartrate object
-     * @throws EasyPostException when the request fails.
-     * @deprecated Use {@link #findLowestSmartrate(List, int, SmartrateAccuracy)} instead.
-     */
-    @Deprecated
-    public static Smartrate findLowestSmartrate(final List<Smartrate> smartrates, int deliveryDay,
-                                                String deliveryAccuracy) throws EasyPostException {
         return findLowestSmartrate(smartrates, deliveryDay, SmartrateAccuracy.getByKeyName(deliveryAccuracy));
     }
 
--- a/src/main/java/com/easypost/model/Shipment.java
+++ b/src/main/java/com/easypost/model/Shipment.java
@@ -3,9 +3,7 @@ package com.easypost.model;
 import com.easypost.exception.EasyPostException;
 import com.easypost.net.EasyPostResource;
 
-import java.util.Arrays;
 import java.util.HashMap;
-import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


src/main/java/com/easypost/model/Shipment.java
```diff
@@ -1035,8 +1033,22 @@ public Shipment insure(final String apiKey) throws EasyPostException {
      * @param deliveryAccuracy Delivery days accuracy restriction to use when filtering.
      * @return lowest Smartrate object
      * @throws EasyPostException when the request fails.
+     * @deprecated use {@link #lowestSmartRate(int, SmartrateAccuracy)} instead.
```
Justintime50, line 1036: Issue: I believe this is missing a param?

src/main/java/com/easypost/model/Shipment.java
```diff
@@ -1090,21 +1102,29 @@ public static Smartrate getLowestSmartRate(final List<Smartrate> smartrates, int
      * @param deliveryAccuracy Delivery days accuracy restriction to use when filtering.
      * @return lowest Smartrate object
      * @throws EasyPostException when the request fails.
+     * @deprecated Use {@link #findLowestSmartrate(List, int, SmartrateAccuracy)} instead.
```
Justintime50, line 1105: Question: Shouldn't these have the names of the params? How does `@link` work?

Justintime50 (commented):
Let's ask a genuine question here: "what does this refactor accomplish that the original did not"? If we are refactoring for the sake of refactoring, we shouldn't do it. I'd like to point out that this takes 75+ extra lines of code to ultimately do the same thing as the current implementation. Enums are good but at what cost if (at least this case) they introduce so much extra code that does the same thing while deprecating a function for the sake of having the ability to use an enum.

Thoughts?

src/main/java/com/easypost/model/Shipment.java
nwithan8: https://stackoverflow.com/questions/11458398/what-does-link-mean-in-java-android-documentation

src/main/java/com/easypost/model/Shipment.java
nwithan8: No, this redirects users to use `lowestSmartRate` on line 1051.

nwithan8:
> Let's ask a genuine question here: "what does this refactor accomplish that the original did not"? If we are refactoring for the sake of refactoring, we shouldn't do it. I'd like to point out that this takes ~75 extra lines of code to ultimately do the same thing as the current implementation. Enums are good but at what cost if (at least this case) they introduce so much extra code that does the same thing while deprecating a function for the sake of having the ability to use an enum.
> 
> Thoughts?

The main purpose of enums is for when there are a limited number of options available for the end user to use, such as here where there are only a handful of valid smartrate accuracies to choose from.
Inherently, using enums prevents errors. Compared to hardcoded strings, they prevent errors due to accidental misspellings, inconsistencies in casing, etc. Plus, those using code completion can easily pull up the list of available enums to them, rather than having to refer to our documentation to find the exact string to use in a specific function.
There's also a [variety of underlying benefits](https://crunchify.com/why-and-for-what-should-i-use-enum-java-enum-examples/) to using enums rather than int-to-int or string-to-string comparisons in terms of compilation, functionality and safety.

Justintime50 (dismissed):
A well defended stance on enums 🙂 I can accept that.

src/main/java/com/easypost/model/TimeInTransit.java
```diff
@@ -152,7 +152,9 @@ public void setPercentile99(final Integer percentile99) {
      * @param percentile the percentile to find the corresponding accuracy for
      * @return the delivery accuracy of the specified percentile
      * @throws EasyPostException when the percentile is not valid
+     * @deprecated Use {@link #getBySmartrateAccuracy(SmartrateAccuracy)} instead.
      */
+    @Deprecated
     public int getSmartRateAccuracy(final String percentile) throws EasyPostException {
```
jchen293, line 158: I believe this function can be removed unless a user is using it for some reason

nwithan8: I'll go through in a second and remove any function that hasn't been released publicly yet.

Welcome to Java, where everything is deprecated, but nothing is removed.

Justintime50 (approved):

jchen293 (approved):

PR #153, opened 2022-06-30
Adds OS specific details to the user-agent header

jchen293:
# Description

Adds OS-specific details to the user-agent header

# Testing

Screen Shot 2022-06-30 at 1 44 55 PM

# Pull Request Type

### DIFF
--- a/src/test/java/com/easypost/TestUtils.java
+++ b/src/test/java/com/easypost/TestUtils.java
@@ -18,12 +18,11 @@ public abstract class TestUtils {
         PRODUCTION
     }
 
-    private static final String API_KEY_FAILED_TO_PULL = "couldnotpullapikey";
+    private static final String API_KEY_FAILED_TO_PULL = "couldNotPullApiKey";
     private static final String CASSETTES_PATH = "src/test/cassettes";
     private static final java.util.List<String> HEADER_CENSORS = new ArrayList<String>() {{
         add("Authorization");
         add("User-Agent");
-        add("X-Client-User-Agent");
     }};
     private static final List<String> QUERY_CENSORS = new ArrayList<String>() {{
         add("card[cvc]");
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   functions.
     - `getSmartrates()` -> `smartrates()`
     - `getLowestSmartRate()` -> `findLowestSmartrate()`
+- Adds OS specific details to the user-agent header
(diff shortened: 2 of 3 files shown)


Justintime50 (changes requested):
The code looks fine, I would however like an `Implementation` piece included like we did with Python. What kind of Java are they using.

jchen293:
> The code looks fine, I would however like an `Implementation` piece included like we did with Python. What kind of Java are they using.

added

jchen293:
> 

This also applies to other sections as well, we need to add this in other libs or the one(s) that already rolled out.

jchen293:
> So the way that our User-Agent header works is that each new "section" is delimited by a space. The `Implementation` section has spaces in it. We may want to join each piece of that section with an underscore or something similar so that the sections are consistent. Does that make sense? (this is based on the new pic)

> 

This also applies to other sections as well, we need to add this in other libs or the one(s) that already rolled out.

Justintime50:
> This also applies to other sections as well, we need to add this in other libs or the one(s) that already rolled out.

Can you supply an example of this happening? I don't believe this is actually the case. In Python, the `Implementation` will be something like `Pypy` or `CPython` and won't have extra spaces in it.

jchen293:
> > This also applies to other sections as well, we need to add this in other libs or the one(s) that already rolled out.
> 
> Can you supply an example of this happening? I don't believe this is actually the case. In Python, the `Implementation` will be something like `Pypy` or `CPython` and won't have extra spaces in it.

Ok, in Python this probably will not be an issue, but for the `OS name`, in my case, it would be `Mac OS X`, we definitely want to convert spaces to hyphens.

src/test/java/com/easypost/TestUtils.java
```diff
@@ -18,12 +18,11 @@ public enum ApiKey {
         PRODUCTION
     }
 
-    private static final String API_KEY_FAILED_TO_PULL = "couldnotpullapikey";
+    private static final String API_KEY_FAILED_TO_PULL = "couldNotPullApiKey";
```
nwithan8, line 21: Why the change?

nwithan8 (approved):

src/test/java/com/easypost/TestUtils.java
jchen293: I prefer the camel case in Java

Justintime50 (approved):

PR #156, opened 2022-07-06
Security analysis and Makefile

nwithan8:
# Description

- Add dependency-check as dependency
- Add Makefile for common commands (lint, security check, build, release, publish, etc)
- Add CI step (see notes)

# Testing

- All Makefile commands work as expected
- Dependency check runs from command line as expected
- All unit tests pass as expected, no changes or re-recording needed

# Notes
- The dependency check generates an HTML report after running, which is then uploaded as an artifact (example below). The CI step itself should fail on severe CVSs (plugin is configured to fail a build if there are any CVSs with a score 7+)
- There is also a [free GitHub integration](https://github.com/marketplace/muse-dev/plan/MLP_kgDNGN8) from Sonatype (who run the Maven repository) to scan for CVSs, that we might want to consider. This should be in addition to the CI tool, not as a replacement, to allow local CVS scanning still
- Example report: image

# Pull Request Type

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -262,6 +262,10 @@
                 <groupId>org.owasp</groupId>
                 <artifactId>dependency-check-maven</artifactId>
                 <version>7.1.1</version>
+                <configuration>
+                    <failBuildOnCVSS>7</failBuildOnCVSS>
+                    <junitFailOnCVSS>7</junitFailOnCVSS>
+                </configuration>
                 <executions>
                     <execution>
                         <goals>
--- a/pom.xml
+++ b/pom.xml
@@ -60,6 +60,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <version>23.0.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>com.easypost</groupId>
             <artifactId>easyvcr</artifactId>
             <version>0.3.0</version>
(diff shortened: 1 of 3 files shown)


Justintime50 (changes requested):
If this tool isn't capable of throwing errors then we should investigate it's usage more or use another tool as a manual tool is counterproductive. 

Additionally, I know this library has security issues (as I remember your IDE getting upset at us some months ago), certainly it should be complaining and there should be some fixes required here (bumping patch versions of deps and whatnot). Can we investigate this a bit more?

nwithan8:
> If this tool isn't capable of throwing errors then we should investigate it's usage more or use another tool as a manual tool is counterproductive.
> 
> Additionally, I know this library has security issues (as I remember your IDE getting upset at us some months ago), certainly it should be complaining and there should be some fixes required here (bumping patch versions of deps and whatnot). Can we investigate this a bit more?

After running Dependency Check a half-dozen different ways with various strict configurations, it has not detected any vulnerabilites.

I know in the past (and still currently), my IntelliJ IDE's Maven tool is reporting that `EasyVCR` is transitively vulnerable to some low-threat CVEs, but Sonatype themselves [do not report any vulnerabilities](https://ossindex.sonatype.org/component/pkg:maven/com.easypost/easyvcr@0.3.0?utm_source=dependency-check&utm_medium=integration&utm_content=7.1.1) with `EasyVCR`, leading me to believe my IDE is wrong.

FWIW, the IDE vulnerability checks are powered by [Checkmarx](https://checkmarx.com/), which is a paid product.

pom.xml
```diff
@@ -59,6 +59,12 @@
             <version>2.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jetbrains</groupId>
```
nwithan8, line 63: This is a dependency for DependencyCheck.

Justintime50 (approved):

jchen293 (approved):

PR #166, opened 2022-07-25
Add carbon offset functionality

nwithan8:
# Description

- Add carbon offset to rate objects
- Add ability to declare carbon offset when creating, buying or rerating a shipment
- Add unit tests, some disabled currently

NOTE: Why do `Rate`, `SmartRate` and `Event` have their own specific GSON serializers?

# Testing

- Added four unit tests

# Pull Request Type


src/main/java/com/easypost/model/RateDeserializer.java
```diff
@@ -131,6 +139,8 @@ public Rate deserialize(final JsonElement json, final Type typeOfT, final JsonDe
         rate.setEstDeliveryDays(estDeliveryDays);
         rate.setShipmentId(shipmentID);
         rate.setCarrierAccountId(jo.get("carrier_account_id").getAsString());
+        rate.setCarbonOffset(carbonOffset);
+        rate.setBillingType(jo.get("billing_type").getAsString());
```
nwithan8, line 143: DEVELOPER NOTE: Went ahead and fixed `billing_type`, was not being serialized as well. Must have slipped through the cracks a long time ago.

src/test/java/com/easypost/CarbonOffsetTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+        Shipment shipment = Shipment.create(Fixture.basicCarbonOffsetShipment(), true);
+
+        assertInstanceOf(Shipment.class, shipment);
+
+        List<Rate> rates = shipment.getRates();
+        assertNotNull(rates);
+
+        Rate rate = rates.get(0);
+        assertNotNull(rate.getCarbonOffset());
```
Justintime50, line 47: Suggestion: This assertion has no use and should be removed since the one below it assumes its set meaning by only including the assertion on line 48, you get this one for free without the need to include it.

src/test/java/com/easypost/CarbonOffsetTest.java
```diff
@@ -0,0 +1,133 @@
...
+            if (fee.getType().equals("CarbonOffsetFee")) {
+                foundCarbonOffset = true;
+                break;
+            }
+        }
+        assertTrue(foundCarbonOffset);
+    }
+
+    /**
+     * Test buying a shipment with a carbon offset.
```
Justintime50, line 82:
```suggestion
     * Test one-call buying a shipment with a carbon offset.
```

src/test/java/com/easypost/CarbonOffsetTest.java
```diff
@@ -0,0 +1,133 @@
...
+        Shipment shipmentWithNewRates = shipment.newRates(true);
+
+        List<Rate> newRates = shipmentWithNewRates.getRates();
+        assertNotNull(newRates);
+
+        Rate oldRate = rates.get(0);
+        Rate newRate = newRates.get(0);
+
+        assertNull(oldRate.getCarbonOffset());
+        assertNotNull(newRate.getCarbonOffset());
```
Justintime50, line 131: Thought: Smart!

src/test/java/com/easypost/Fixture.java
```diff
@@ -431,4 +431,60 @@ public static Map<String, Object> rmaFormOptions() {
...
+        carrierAccounts.add(uspsCarrierAccountID());
+        oneCallBuyShipment.put("to_address", pickupAddress());
+        oneCallBuyShipment.put("from_address", basicAddress());
+        oneCallBuyShipment.put("parcel", basicParcel());
+        oneCallBuyShipment.put("service", uspsService());
+        oneCallBuyShipment.put("carrier_accounts", carrierAccounts);
+        oneCallBuyShipment.put("carrier", usps());
+
+        return oneCallBuyShipment;
+    }
```
Justintime50, line 489: Issue: Please remove all of these `carbon offset shipment` fixtures as they are identical to the already existing shipment fixtures with a new name. Simply call the `basic`, `full`, or `one-call-buy` shipment fixtures in your tests instead. The point of fixtures is to provide as few data points as possible that can be reused as much as possible and mixed/matched as needed for test data.

Justintime50 (changes requested):
I hate overloading in Java... so much code. The code itself seems fine, a few minor things need correcting. 

There appears to be more cassettes re-recorded than is necessary. The only cassettes that should have been changed are shipment ones. Can you identify why the others either were re-recorded or complained that they no longer matched?

nwithan8:
> I hate overloading in Java... so much code. The code itself seems fine, a few minor things need correcting.
> 
> There appears to be more cassettes re-recorded than is necessary. The only cassettes that should have been changed are shipment ones. Can you identify why the others either were re-recorded or complained that they no longer matched?

"carbon_offset" is always included in the shipment creation JSON body now (set to either "true" or "false"). This changed the body of all previously-recorded shipment creation API calls, requiring re-recording of several unit tests.

src/test/java/com/easypost/Fixture.java
nwithan8: These are different, slightly. Normal shipment fixtures have the same `to` and `from` address (the `BasicAddress` fixture). For carbon offset to work, the `to` and `from` addresses need to be different. So these shipments have the `BasicAddress` as the `from` address, and the `PickupAddress` as the `to` address.

I'm not a huge fan of this either. Please advise.

Justintime50:
> "carbon_offset" is always included in the shipment creation JSON body now (set to either "true" or "false"). This changed the body of all previously-recorded shipment creation API calls, requiring re-recording of several unit tests.

If that's the case, that's not correct. If the user doesn't explicitly say they do or don't want it, we should not be passing that key at all (it should be null)

src/test/java/com/easypost/Fixture.java
Justintime50: Interesting. I wondered when this would come to bite us. My suggestion instead is to simply add another basic address fixture so that these are always different. I can handle that separately next sprint when I consolidate all the fixture data. This is fine for now, but it'll get changed in the future.

Justintime50 (dismissed):

Justintime50 (approved):

PR #173, opened 2022-08-03
Skip deprecated functions when merging objects

nwithan8:
# Description

We noticed that, despite it being deprecated, the old `getSmartrates` function is still being called when updating a shipment, as our merge function believes it is a getter. This results in erroneous API calls.

The old function is slated for removal "eventually", but until then, we need to skip the function.

Rather than hard-coding a skip for this specific function, we have added an if clause to the merge functionality that checks if a function is marked as deprecated. If it is, the function is skipped.

# Testing

- Re-recorded two cassettes that had the bad API call in them. Confirmed that bad API call is no longer being made after fix.

# Pull Request Type

### DIFF
--- a/src/main/java/com/easypost/net/EasyPostResource.java
+++ b/src/main/java/com/easypost/net/EasyPostResource.java
@@ -785,6 +785,11 @@ public abstract class EasyPostResource {
         for (Method fromMethod : methods) {
             if (fromMethod.getName().startsWith("get") || GLOBAL_FIELD_ACCESSORS.contains(fromMethod.getName())) {
 
+                if (fromMethod.isAnnotationPresent(Deprecated.class)) {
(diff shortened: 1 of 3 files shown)


Justintime50 (approved):
Question: Why skip deprecated methods altogether? Couldn't this have unintended side-effects vs hard-coding the known one bad call?

nwithan8:
> Question: Why skip deprecated methods altogether? Couldn't this have unintended side-effects vs hard-coding the known one bad call?

So this doesn't happen again where we might not notice it or catch it.

If we've marked a function as deprecated, we don't want people to use it. It also means WE don't want to use it; we usually rewrite and adjust as much as possible to avoid a deprecated function when we do mark something as deprecated.

jchen293 (approved):

PR #174, opened 2022-08-03
Consolidate java-templates folder into rest of module

nwithan8:
# Description

- Move `EasyPost.java` into module rather than as a generated source
- Mark immutable variables as final
- Remove unused `apiVersion` global variable

# Testing

# Pull Request Type


src/main/java/com/easypost/EasyPost.java
```diff
@@ -4,10 +4,9 @@
 
 public abstract class EasyPost {
     public static final String VERSION = "${project.version}";
-    public static String API_BASE = "https://api.easypost.com/v2";
-    public static String BETA_API_BASE = "https://api.easypost.com/beta";
+    public static final String API_BASE = "https://api.easypost.com/v2";
+    public static final String BETA_API_BASE = "https://api.easypost.com/beta";
     public static String apiKey;
-    public static String apiVersion;
```
jchen293, line 10: why is this removed?

nwithan8: Wasn't being used anywhere. Might have been something that snuck in pre-emptively a while ago in anticipation of API versioning.

jchen293 (dismissed):

Justintime50 (commented):
I assume that Java magically knows where to find these values even though the location of the file changed?

Does this file have anything to do with the `templating-maven-plugin` we use in this repo by chance?

nwithan8:
> I assume that Java magically knows where to find these values even though the location of the file changed?
> 
> Does this file have anything to do with the `templating-maven-plugin` we use in this repo by chance?

In Java, the file path doesn't really matter, it's the namespace.

However, after more investigation, it seems we need to keep this template file. The [Templating Maven Plugin ](https://www.mojohaus.org/templating-maven-plugin/) compiles any template files (like this one) at compile time, extracting template information (in this case, the `project.version` value from the `pom.xml` file). If we move this file, it will no longer be considered a template, and the version number will not be extracted properly.

Justintime50:
Aha. Can we please instead add a comment to the file explaining its purpose and how it works with the plugin so future travelers know? That plus the removal of the `apiVersion` var should be good enough.

Justintime50 (approved):
Heh nice

Justintime50:
@jchen293 let's review this ASAP. I want to cut a release of the Java lib with this included so we can get that `getSmartrates` fix out.

jchen293 (approved):

PR #180, opened 2022-08-25
Use examples fixtures for unit tests

nwithan8:
# Description

- New utilities to convert fixtures JSON from examples submodule into usable HashMaps and data for unit tests
- New custom JSON serialization rules for serializing fixtures for API calls
  - As a result, the Fixtures utility file exists in main namespace rather than test namespace since it needs to be referenced by the custom serializer
- Update all unit tests and re-record all cassettes

# Testing

- Re-recorded all cassettes, all unit tests pass

# Pull Request Type


nwithan8:
Got so caught up in fixing the JSON serialization that, oops, forgot to have the CI import the submodule that has the data in it in the first place.

src/main/java/com/easypost/utils/Fixture.java
```diff
@@ -0,0 +1,390 @@
...
+ */
+public abstract class Fixture {
+
+    private static String readFixtureData() {
+        Path fixtureDataPath =
+                Paths.get(getSourceFileDirectory(), "examples/official/fixtures/client-library-fixtures.json");
+        return readFile(fixtureDataPath);
+    }
+
+    private static HashMap<String, Object> createFixture(String data) {
```
Justintime50, line 34: Question: This is all very sad we need so many utils and can't just read from the file. Why do we need so many tools? Can we not do what we did in Go and coerce json blobs to classes?

src/main/java/com/easypost/utils/Fixture.java
```diff
@@ -0,0 +1,390 @@
...
+     *
+     * @return The default pickup date
+     */
+    public static String pickupDate() {
+        /*
+        If you need to re-record cassettes, increment the date below and ensure it is one day in the future,
+        USPS only does "next-day" pickups including Saturday but not Sunday or Holidays.
+         */
+        return "2022-08-30";
+    }
```
Justintime50, line 159: Nit: would prefer this just be in the `pickup` fixture like other libs so someone doesn't need to go finding this since it will be the only one differing from this pattern.

README.md
```diff
@@ -90,6 +90,10 @@ Upgrading major versions of this project? Refer to the [Upgrade Guide](UPGRADE_G
 ### Tests
 
 ```bash
+# Update submodules
+git submodule init
+git submodule update --remote
```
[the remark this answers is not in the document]
nwithan8: Sounds good. I put them in that order since you would need to do that step before running tests.

Makefile
```diff
@@ -18,6 +18,11 @@ coverage:
 install-checkstyle:
 	wget -O checkstyle.jar -q https://github.com/checkstyle/checkstyle/releases/download/checkstyle-10.3.1/checkstyle-10.3.1-all.jar
 
+## install-dev - Install dev requirements
+install-dev:
```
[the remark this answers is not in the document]
nwithan8: Resolved.

src/main/java/com/easypost/utils/Fixture.java
nwithan8: Resolved.

src/test/java/com/easypost/beta/ReferralTest.java
```diff
@@ -22,8 +21,6 @@
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public final class ReferralTest {
-    private static final String REFERRAL_USER_PROD_API_KEY =
```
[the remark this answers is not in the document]
nwithan8: `REFERRAL` is a type of API key that can be pulled from the local environment, just like `TEST` and `PRODUCTION`. And just like them, it has a fallback if the key cannot be pulled from the environment successfully:

image

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -58,6 +59,8 @@ protected enum RequestMethod {
     public static final String EASYPOST_SUPPORT_EMAIL = "support@easypost.com";
     public static final Gson GSON =
             new GsonBuilder().setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
+                    .serializeNulls()
```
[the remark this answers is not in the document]
nwithan8: Removing. Ultimately unnecessary since the custom HashMap serialization rules override this behavior anyway.

As for users, it won't have any impact. This is for serializing (taking params and turning them into JSON data), not deserializing, so it won't impact the final Java objects/models.

src/main/java/com/easypost/net/HashMapSerializer.java
```diff
@@ -0,0 +1,177 @@
+package com.easypost.net;
```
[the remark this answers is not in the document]
nwithan8: Unfortunately not. The `main` namespace isn't aware of the `test` namespace (the `test` namespace knows about the `main` namespace), so I can't host any files that the `main` namespace needs in the `test` namespace.

After working on this, I can actually see the bright side of this work, that being making the JSON serialization more visible. As I note in the code comment, since this is an override for how Gson should serialize any HashMaps, and virtually all data that will need to be serialized for our API calls are HashMaps, this ruleset will handle the majority of all JSON serialization for our application. Previously, this process was blackboxed into Gson's internal `toJsonTree` method, with little control for us to override defaults or troubleshoot invalid serializations. Now that we have effectively overridden that functionality with our own, we have more control over the process if it were to ever break or need to be altered. This also helps us ensure the process is more uniform and standardized.

The C# library underwent a similar renovation a few months ago, where we implemented a custom JSON serializer of sorts, allowing us greater control and access to the process.

The Java library's JSON process has been a sticky operation for some time now (we've had multiple internal tickets in an attempt to fix it or redesign it). I believe this is a step in the right direction.

It has been tested against every API call in our library currently and correctly serialized all data. At the current moment, it seems to cover every case (the HashMap serializer ruleset is quite similar to the [censoring ruleset](https://github.com/EasyPost/easyvcr-java/blob/66c96da9415d8d8889a8dc342deda19e717f22d9/src/main/java/com/easypost/easyvcr/Censors.java#L88) built for the EasyVCR C#/Java applications, if you want more evidence that this implementation is effective).

src/main/java/com/easypost/utils/Fixture.java
nwithan8: That's similar to what is happening here, in the sense that the large JSON string is being serialized into a giant map, then a sub-map is being extracted at a specific key.

I believe the difference comes where, in Go, all our functions are called with an object (an `Address`, a `Shipment`, etc.) used as the parameter, whereas in Java and C#, the functions take in maps of parameters. So, when it comes to deserializing the fixture JSON, we aren't using specific types for JSON properties, but rather generic map types.

We could potentially create a new model where the properties are different elements from the fixture JSON string (think the same way we make a `Shipment` object, with `Parcel` and `Address` properties, out of a normal JSON string), and then simply extract those properties to use as fixtures. I will try that now, but I'm unsure of the performance or clarity benefits it might bring (plus, read my comment above re: the benefits of the new HashMap serializer)

nwithan8: I have changed how our fixtures are generated programmatically, going with a more explicit option that will is pro: more readable, con: more to maintain, unknown: more stable.

As a result, the fixtures exist again in the `test` namespace. I was also able to move some file-related utilities back to the `test` namespace, and remove fixture-related methods from the HashMap serializer.

I am still going to keep the new HashMap serializer for the benefits listed in my comment above, but in terms of loading the fixture JSON for unit tests, this does seem to be the better way to go (even if it undoes all my fancy trickery 🙃 )

Awaiting your feedback.

src/main/java/com/easypost/net/HashMapSerializer.java
```diff
@@ -0,0 +1,87 @@
...
+        for (Map.Entry<String, Object> entry : cleanedHashMap.entrySet()) {
+            String key = entry.getKey();
+            Object value = entry.getValue();
+
+            if (value == null) {
+                // don't serialize nulls
+                continue;
+            }
+
+            jsonObject.add(key, gson.toJsonTree(value));
```
jchen293, line 82: nit:
```suggestion
            if (value != null) {
                jsonObject.add(key, gson.toJsonTree(value));
            }
```

nwithan8: Good suggestion. I prefer the way it is since it's easier to read what happens when value is null. Thoughts @Justintime50 ?

src/test/java/com/easypost/Fixture.java
```diff
@@ -118,400 +57,265 @@ public static String uspsCarrierAccountID() {
...
      */
-    public static Map<String, Object> taxIdentifier() {
-        Map<String, Object> taxIdentifier = new HashMap<>();
+    public static HashMap<String, Object> basicPickup() {
+        HashMap<String, Object> fixture = Objects.requireNonNull(getFixtureData()).pickups.basic;
+
+        /*
+        If you need to re-record cassettes, increment the date below and ensure it is one day in the future,
+        USPS only does "next-day" pickups including Saturday but not Sunday or Holidays.
+         */
```
Justintime50, line 232: Nit: Should really be in the doctoring of the function and not an internal docstring 🤷‍♂️

src/main/java/com/easypost/net/HashMapSerializer.java
Justintime50: Jake's suggestion is cleaner and is how I would have done it myself; however, either work.

Makefile
```diff
@@ -18,6 +18,11 @@ coverage:
 install-checkstyle:
 	wget -O checkstyle.jar -q https://github.com/checkstyle/checkstyle/releases/download/checkstyle-10.3.1/checkstyle-10.3.1-all.jar
 
+## install-dev - Install dev requirements
```
Justintime50, line 21:
```suggestion
## install- Install dev requirements
```

Makefile
```diff
@@ -47,4 +52,4 @@ scan:
 test:
 	mvn --batch-mode install -Dgpg.skip=true -Dcheckstyle.skip=true -Dcheckstyle.skip=true -Ddependency-check.skip=true -Djavadoc.skip=true
 
-.PHONY: help build clean install-checkstyle lint publish publish-dry release scan scan-strict test
+.PHONY: help build clean install-checkstyle install-dev lint publish publish-dry release scan scan-strict test
```
Justintime50, line 55:
```suggestion
.PHONY: help build clean install-checkstyle install lint publish publish-dry release scan scan-strict test
```

Justintime50 (changes requested):
1. Prefer `Fixtures` to `FixtureData` so it matches other libs/expectations 
2. Don't care for the `sections` directory name, can we rename to something that better captures what is contained? Sections of what? Can we instead say `components` or something similar?

src/test/java/com/easypost/Fixture.java
nwithan8: Resolved.

Justintime50 (approved):

src/test/java/com/easypost/fixtures/components/ServiceNamesUsps.java
```diff
@@ -0,0 +1,11 @@
+package com.easypost.fixtures.components;
+
+import com.google.gson.annotations.SerializedName;
+
+public final class ServiceNamesUsps {
+    @SerializedName("first_service")
+    public String firstService;
+
+    @SerializedName("pickup_service")
+    public String pickupService;
```
jchen293, line 10:
```suggestion
    public String pickupService;
    
```
nit: extra line

jchen293 (approved):

PR #182, opened 2022-09-07
fix: checkstyle install target

Justintime50:
# Description

Fixes the install target to install checkstyle too, ignores jar files and sorts gitignore file.

# Testing

# Pull Request Type


nwithan8:
I feel like there was a reason I went with wget rather than curl, but I can't remember. If it work, it works.

nwithan8 (approved):

Justintime50:
> I feel like there was a reason I went with wget rather than curl, but I can't remember. If it work, it works.

`wget` is good for container workflows because it's often installed by default for Alpine Linux distros; however, the opposite is true of normal users where macOS isn't guaranteed to have it whereas it does `curl`. My approach is use the tool that the majority of the people using/maintaining the project are using so it works for the most amount of people out of the box.

jchen293 (approved):
Thanks for fixing this, I had to manually install `wget` on my Mac

nwithan8:
> > I feel like there was a reason I went with wget rather than curl, but I can't remember. If it work, it works.
> 
> `wget` is good for container workflows because it's often installed by default for Alpine Linux distros; however, the opposite is true of normal users where macOS isn't guaranteed to have it whereas it does `curl`. My approach is use the tool that the majority of the people using/maintaining the project are using so it works for the most amount of people out of the box.

Gotcha. Meanwhile, I'm pretty sure I was on my Windows machine when I did this, which comes with neither pre-installed, but I had previously added both to my PATH, so... 🤷🏻

PR #183, opened 2022-09-15
add endshipper id in shipment buy and remove unnecessary buy functions

jchen293:
# Description

Allow user to pass in EndShipper ID in Shipment buy function
Remove unnecessary buy functions that do not have rate parameters
# Testing

E2E test, will try to use Mock testing in the future

# Pull Request Type

### DIFF
--- a/src/main/java/com/easypost/model/Shipment.java
+++ b/src/main/java/com/easypost/model/Shipment.java
@@ -979,18 +979,6 @@ public final class Shipment extends EasyPostResource {
     /**
      * Buy this Shipment.
      *
-     * @param endShipperId the id of the end shipper to use for this purchase.
-     * @param apiKey       API key to use in request (overrides default API key).
-     * @return Shipment object
-     * @throws EasyPostException when the request fails.
-     */
-    public Shipment buy(final String endShipperId, final String apiKey) throws EasyPostException {
-        return this.buy(new HashMap<String, Object>() {}, endShipperId, apiKey);
-    }
-
-    /**
-     * Buy this Shipment.
-     *
      * @param rate the Rate to use for this Shipment.
      * @return Shipment object
      * @throws EasyPostException when the request fails.
@@ -1039,11 +1027,10 @@ public final class Shipment extends EasyPostResource {
      * @param apiKey API key to use in request (overrides default API key).
      * @return Shipment object
      * @throws EasyPostException when the request fails.
-     *
-     * When Java Client Library rewrite happens, the apiKey param will be replaced with endShipperId
      */
     public Shipment buy(final Map<String, Object> params, final String apiKey) throws EasyPostException {
-        return this.buy(params, false, null, apiKey);
+      // TODO: When Java Client Library rewrite happens, the apiKey param will be replaced with endShipperId  
+      return this.buy(params, false, null, apiKey);
     }
 
     /**
@@ -1098,13 +1085,14 @@ public final class Shipment extends EasyPostResource {
      * @return Shipment object
      * @throws EasyPostException when the request fails.
      */
-    public Shipment buy(final Map<String, Object> params, final boolean withCarbonOffset, 
+    public Shipment buy(final Map<String, Object> params, final boolean withCarbonOffset,
         final String endShipperId, final String apiKey) throws EasyPostException {
+        params.put("carbon_offset", withCarbonOffset);
+
         if (endShipperId != null && !endShipperId.isEmpty()) {
-            params.put("end_shipper_id", endShipperId);
+          params.put("end_shipper_id", endShipperId);
         }
 
-        params.put("carbon_offset", withCarbonOffset);
         Shipment response =
                 request(RequestMethod.POST, String.format("%s/buy", instanceURL(Shipment.class, this.getId())), params,
                         Shipment.class, apiKey);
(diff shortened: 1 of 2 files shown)


src/main/java/com/easypost/model/Shipment.java
```diff
@@ -976,6 +976,18 @@ public Shipment buy(final boolean withCarbonOffset, final String apiKey) throws
...
 
+    /**
+     * Buy this Shipment.
+     *
+     * @param endShipperId the id of the end shipper to use for this purchase.
+     * @param apiKey       API key to use in request (overrides default API key).
+     * @return Shipment object
+     * @throws EasyPostException when the request fails.
+     */
+    public Shipment buy(final String endShipperId, final String apiKey) throws EasyPostException {
```
Justintime50, line 987: Issue: We will never be able to use this function because a rate **must** be passed to buy a shipment.

src/main/java/com/easypost/model/Shipment.java
```diff
@@ -1012,9 +1039,25 @@ public Shipment buy(final Rate rate, final boolean withCarbonOffset) throws Easy
      * @param apiKey API key to use in request (overrides default API key).
      * @return Shipment object
      * @throws EasyPostException when the request fails.
+     *
+     * When Java Client Library rewrite happens, the apiKey param will be replaced with endShipperId
```
Justintime50, line 1043: Issue: This must be a TODO comment to signify an action is required from us. This should also be in the code itself and not the docstring since the docstring is viewable by users and this note is only intended for us.

src/main/java/com/easypost/model/Shipment.java
```diff
@@ -1028,6 +1071,39 @@ public Shipment buy(final Map<String, Object> params, final String apiKey) throw
...
+     * @param endShipperId     the id of the end shipper to use for this purchase.
+     * @param apiKey           API key to use in request (overrides default API key).
+     * @return Shipment object
+     * @throws EasyPostException when the request fails.
+     */
+    public Shipment buy(final Map<String, Object> params, final boolean withCarbonOffset, 
+        final String endShipperId, final String apiKey) throws EasyPostException {
+        if (endShipperId != null && !endShipperId.isEmpty()) {
+            params.put("end_shipper_id", endShipperId);
+        }
```
Justintime50, line 1105: Issue: Order of operations here, this should come after carbon offset since that param comes before end shipper.

Justintime50 (changes requested):

src/main/java/com/easypost/model/Shipment.java
jchen293: We need to overhaul `buy` in Shipment, there are a few `buy` functions that do not have `rate` or `param` in the parameters. For example: https://github.com/EasyPost/easypost-java/blob/master/src/main/java/com/easypost/model/Shipment.java#L918-L931

src/main/java/com/easypost/model/Shipment.java
jchen293: fixed

src/main/java/com/easypost/model/Shipment.java
jchen293: fixed

src/main/java/com/easypost/model/Shipment.java
jchen293: fixed, but I believe we need to remove/deprecate other unnecessary functions in a separate PR

Justintime50: Right, we need to remove any override buy function that doesn't contain a `rate` parameter because it will error out at the API because no rate was passed.

Let's not add an unusable override in this PR. Since we are fixing the buys here, you can remove any buy override function that does not have a rate and add a CHANGELOG entry stating we `removed unusable shipment buy functions that did not contain a rate parameter`

jchen293: It's called overload :)

src/main/java/com/easypost/model/Shipment.java
```diff
@@ -912,123 +912,135 @@ public List<Smartrate> smartrates(final String apiKey) throws EasyPostException
...
      * @return Shipment object
      * @throws EasyPostException when the request fails.
      */
-    public Shipment buy(final Map<String, Object> params, final boolean withCarbonOffset, final String apiKey)
-            throws EasyPostException {
+    public Shipment buy(final Map<String, Object> params, final boolean withCarbonOffset,
+        final String endShipperId, final String apiKey) throws EasyPostException {
         params.put("carbon_offset", withCarbonOffset);
+
+        if (endShipperId != null && !endShipperId.isEmpty()) {
```
nwithan8, line 1040: 11 overloads all to boil down to one actual implementation, I love Java 😄

nwithan8 (approved):

src/main/java/com/easypost/model/Shipment.java
```diff
@@ -941,41 +920,6 @@ public Shipment buy(final Map<String, Object> params) throws EasyPostException {
...
-     * Buy this Shipment.
-     *
-     * @param params           the options for the query.
-     * @param withCarbonOffset whether to include a carbon offset when buying the shipment.
-     * @return Shipment object
-     * @throws EasyPostException when the request fails.
-     */
-    public Shipment buy(final Map<String, Object> params, final boolean withCarbonOffset) throws EasyPostException {
-        return this.buy(params, withCarbonOffset, null);
-    }
```
Justintime50, line 954: Issue: This function must remain because users could pass the rate via the `params` parameter.

Justintime50 (changes requested):

Justintime50 (approved):

PR #184, opened 2022-09-16
Move Partner to GA

nwithan8:
# Description

- Copy Referral classes to GA, remove deprecated functions that don't need to carry over
- Replace beta endpoints with GA endpoints
- Mark beta Referral classes as deprecated (NOTE: When deleting, can probably delete deprecated CreditCard classes too)
- Copy Referral unit tests to GA, keep beta unit tests as well
- Add missing README.md info
- Add missing Partner key to TestUtils

# Testing

- Tests re-recorded as needed, all pass

# Pull Request Type


src/main/java/com/easypost/model/ReferralCustomer.java
```diff
@@ -0,0 +1,257 @@
...
+     * @param params Map of the referral user parameters.
+     * @return Referral object.
+     * @throws EasyPostException when the request fails.
+     */
+    public static ReferralCustomer create(Map<String, Object> params) throws EasyPostException {
+        return create(params, null);
+    }
+
+    /**
+     * Create a Referral object from parameter map. This function requires the Partner User's API key.
```
jchen293, line 49:
```suggestion
     * Create a Referral Customer object from parameter map. This function requires the Partner User's API key.
```

jchen293 (approved):

Makefile
```diff
@@ -6,6 +6,10 @@ help:
 build:
 	mvn clean install -DskipTests=true -Dgpg.skip=true -Dcheckstyle.skip=true -Dcheckstyle.skip=true -Ddependency-check.skip=true -Djavadoc.skip=true
 
+## build-deps - Build the project dependencies
+build-deps:
+	mvn install
```
Justintime50, line 11: Question: Shouldn't this instead be in the `install` target instead of its own like all the other libs?

src/main/java/com/easypost/model/beta/ReferralCustomer.java
```diff
@@ -19,6 +19,10 @@
 import java.util.List;
 import java.util.Map;
 
+/**
+ * @deprecated Use {@link com.easypost.model.ReferralCustomer} instead.
+ * Deprecated: v5.10.0 - v7.0.0
```
Justintime50, line 24: Issue: for beta features, we won't use the traditional "2 version removal" policy, since we can't guarantee the beta feature itself at the API level will live on that long. Once the API beta endpoint is gone, we are free to remove from the lib since it will become unusable. For that reason, let's not tag these with the version ranges.

If this was a deprecated non-beta feature, than the tags would absolutely apply.

src/main/java/com/easypost/model/beta/ReferralCustomerCollection.java
```diff
@@ -4,6 +4,10 @@
 
 import java.util.List;
 
+/**
+ * @deprecated Use {@link com.easypost.model.ReferralCustomerCollection} instead.
+ * Deprecated: v5.10.0 - v7.0.0
```
Justintime50, line 9: Issue: for beta features, we won't use the traditional "2 version removal" policy, since we can't guarantee the beta feature itself at the API level will live on that long. Once the API beta endpoint is gone, we are free to remove from the lib since it will become unusable. For that reason, let's not tag these with the version ranges.

If this was a deprecated non-beta feature, than the tags would absolutely apply.

Makefile
nwithan8: When it's in "install", it's run as part of CI.
There (and locally), it throws an error due to the GPG plugin configuration.
Separating this out into a separate step (one that doesn't need to be run manually in most cases unless something goes wrong) is, IMO, a better backup.
For reference, this is the solution to the templating situation (explained in the wall of text in `src/main/java-templates/com/easypost/EasyPost.java`

nwithan8: Resolved

src/main/java/com/easypost/model/beta/ReferralCustomer.java
nwithan8: Resolved

src/main/java/com/easypost/model/beta/ReferralCustomerCollection.java
nwithan8: Resolved

Justintime50 (dismissed):

Justintime50 (approved):

PR #193, opened 2022-10-17
Improve error handling

jchen293:
# Description

- Introduce a handful of explicit error types based on the status code
- Two types of error: API or General
   - API: API exceptions are generated based on the status code of the failed API call.
   - General: Exceptions from client library such as no lowest rate found.
- Improve the error parsing when the error message is an array(rare case) by concatenating the array to a string.

# Testing

Add unit tests to make all tests pass

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,11 +17,6 @@
 - Improves error exception handling
   - Specific error types for each category of error
   - API error message may be an array rather than a string. Arrays will be concatenated (by comma) and returned as a string.
-- Adds two missing attributes in the Event class: `pendingUrls` and `completedUrls`
-- Changes the type `result` of Event from `EasyPostResource` to `Map<String, Object>`
-- Improves error exception handling
-  - Specific error types for each category of error
-  - API error message may be an array rather than a string. Arrays will be concatenated (by comma) and returned as a string.
 
 ## v5.10.0 (2022-09-21)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,16 @@
 - Changes the type of Insurance `Amount` from Float to String
 - Removes `getUpdate` and `setUpdate` from Tracker class
 - Removes all beta features and corrsponding unit tests
-- Remove `serviceCode` from `Rate` class since this value is internal use only
+- Removes `serviceCode` from `Rate` class since this value is internal use only
 - Converts return type from boolean to void in empty response body functions
   - `fundWallet()` and `deletePaymentMethod()` in Billing class
   - `createList()` in Tracker class
   - `updateEmail()` in ReferralCustomer class
 - Adds two missing attributes in the Event class: `pendingUrls` and `completedUrls`
 - Changes the type `result` of Event from `EasyPostResource` to `Map<String, Object>`
+- Improves error exception handling
+  - Specific error types for each category of error
+  - API error message may be an array rather than a string. Arrays will be concatenated (by comma) and returned as a string.
 
 ## v5.10.0 (2022-09-21)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@
 - Improves error exception handling
   - Specific error types for each category of error
   - API error message may be an array rather than a string. Arrays will be concatenated (by comma) and returned as a string.
+- Adds two missing attributes in the Event class: `pendingUrls` and `completedUrls`
+- Changes the type `result` of Event from `EasyPostResource` to `Map<String, Object>`
 
 ## v5.10.0 (2022-09-21)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@
   - API error message may be an array rather than a string. Arrays will be concatenated (by comma) and returned as a string.
 - Adds two missing attributes in the Event class: `pendingUrls` and `completedUrls`
 - Changes the type `result` of Event from `EasyPostResource` to `Map<String, Object>`
+- Improves error exception handling
+  - Specific error types for each category of error
+  - API error message may be an array rather than a string. Arrays will be concatenated (by comma) and returned as a string.
 
 ## v5.10.0 (2022-09-21)
 
--- a/pom.xml
+++ b/pom.xml
@@ -64,6 +64,11 @@
             <version>0.4.2</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+            <version>4.5.13</version>
+        </dependency>
     </dependencies>
 
     <organization>
--- a/src/main/java/com/easypost/exception/EasyPostException.java
+++ b/src/main/java/com/easypost/exception/EasyPostException.java
@@ -11,47 +11,73 @@ package com.easypost.exception;
 public class EasyPostException extends Exception {
 
     private static final long serialVersionUID = 1L;
-    private final String param;
+    private final String code;
+    private final Integer statusCode;
+    private final String message;
 
     /**
-     * Get the parameter name.
-     *
-     * @return the parameter name
-     */
-    public String getParam() {
-        return param;
-    }
-
-    /**
-     * Constructor.
+     * EasyPostException constructor.
      *
      * @param message the exception message
      */
     public EasyPostException(final String message) {
-        super(message, null);
-        this.param = null;
+        super(message);
+        this.code = null;
+        this.statusCode = null;
+        this.message = message;
     }
 
     /**
-     * Constructor.
+     * EasyPostException constructor.
      *
      * @param message the exception message
      * @param ex      the exception cause
      */
     public EasyPostException(final String message, final Throwable ex) {
         super(message, ex);
-        this.param = null;
+        this.code = null;
+        this.statusCode = null;
+        this.message = message;
     }
 
     /**
-     * Constructor.
+     * EasyPostException constructor.
      *
      * @param message the exception message
-     * @param param   the parameter name
-     * @param ex      the exception cause
+     * @param code the exception code
+     * @param statusCode the exception status code
      */
-    public EasyPostException(final String message, final String param, final Throwable ex) {
-        super(message, ex);
-        this.param = param;
+    public EasyPostException(final String message, final String code, final int statusCode) {
+        super(message);
+        this.code = code;
+        this.statusCode = statusCode;
+        this.message = message;
+    }
+
+    /**
+     * Get status code of the error object.
+     * 
+     * @return statusCode the status code of the error object
+     */
+    public Integer getStatusCode() {
+        return statusCode;
+    }
+
+    /**
+     * Get code of the error object.
+     *
+     * @return code the code of the error object
+     */
+    public String getCode() {
+        return code;
+    }
+
+    /**
+     * Get message of the error object.
+     *
+     * @return message the message of the error object
+     */
+    public String getMessage() {
+        return message;
     }
 }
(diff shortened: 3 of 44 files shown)


src/main/java/com/easypost/model/Error.java
```diff
@@ -1,45 +1,46 @@
...
-        return code;
+    public List<Error> getError() {
+        return errors;
     }
 
     /**
-     * Set the error code.
-     *
-     * @param code code
+     * set the errors for an Error object.
```
Justintime50, line 20:
```suggestion
     * Set the errors for an Error object.
```

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -9,7 +9,24 @@
...
+import com.easypost.exception.API.InvalidRequestError;
+import com.easypost.exception.API.MethodNotAllowedError;
+import com.easypost.exception.API.NotFoundError;
+import com.easypost.exception.API.PaymentError;
+import com.easypost.exception.API.RateLimitError;
+import com.easypost.exception.API.RedirectError;
+import com.easypost.exception.API.ServiceUnavailablError;
+import com.easypost.exception.API.TimeoutError;
+import com.easypost.exception.API.UnauthorizedError;
+import com.easypost.exception.API.UnknownApiError;
```
Justintime50, line 26: Can we not just import `com.easypost.exception.API.*`?

src/test/java/com/easypost/ErrorTest.java
```diff
@@ -1,13 +1,33 @@
...
+import com.easypost.exception.API.UnknownApiError;
+import com.easypost.exception.API.PaymentError;
+import com.easypost.exception.API.RateLimitError;
+import com.easypost.exception.API.NotFoundError;
+import com.easypost.exception.API.MethodNotAllowedError;
+import com.easypost.exception.API.TimeoutError;
+import com.easypost.exception.API.ForbiddenError;
+import com.easypost.exception.API.GatewayTimeoutError;
+import com.easypost.exception.API.InternalServerError;
+import com.easypost.exception.API.InvalidRequestError;
```
Justintime50, line 18: Same import question as above?

src/test/java/com/easypost/ErrorTest.java
```diff
@@ -34,4 +54,77 @@ public void testError() throws EasyPostException {
         // so we'll just check fo a generic exception
         assertThrows(Exception.class, () -> Shipment.create(null));
     }
+
+    /**
+     * Test every error type and make sure each errro has property.
```
Justintime50, line 59:
```suggestion
     * Test every error type and make sure each error has the correct properties.
```

Justintime50 (commented):
This looks great, nice work! There were a couple tiny spelling things to fix otherwise.

Let's please add one more test, a real API call to test that we throw errors correctly for invalid address verification. You can check any of the other test suites `address` tests files for an example on how to replicate this.

src/main/java/com/easypost/net/EasyPostResource.java
jchen293: Yes, we could, but the linter will complain. The only issue with using wildcard import is it could clutter your local namespace. More details here: https://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad

src/test/java/com/easypost/ErrorTest.java
jchen293: Look my previous answer

src/main/java/com/easypost/net/EasyPostResource.java
Justintime50: I'm aware, this was the same question I asked in your interview 😉 

If the linter complains, then feel free to leave it. The reason why a wildcard error shouldn't matter much here is because we have complete control over the environment and don't really use other dependencies meaning the namespace collision possibility would be super tiny since we're only importing our own items.

nwithan8: But it opens up others to potentially back-door accessing other classes (think a bad __init__ in Python)

pom.xml
```diff
@@ -64,6 +64,11 @@
             <version>0.4.2</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
```
jchen293, line 67: ## Reviewer note:

Adding this dependency because we were using the dependency from the `EasyVCR` pom.xml, we should explicit this out in the CL itself.

CHANGELOG.md
```diff
@@ -7,13 +7,21 @@
...
 - Adds two missing attributes in the Event class: `pendingUrls` and `completedUrls`
 - Changes the type `result` of Event from `EasyPostResource` to `Map<String, Object>`
+- Improves error exception handling
+  - Specific error types for each category of error
+  - API error message may be an array rather than a string. Arrays will be concatenated (by comma) and returned as a string.
+- Adds two missing attributes in the Event class: `pendingUrls` and `completedUrls`
+- Changes the type `result` of Event from `EasyPostResource` to `Map<String, Object>`
+- Improves error exception handling
+  - Specific error types for each category of error
+  - API error message may be an array rather than a string. Arrays will be concatenated (by comma) and returned as a string.
```
Justintime50, line 24: Issue: you have this in here twice

pom.xml
Justintime50: Question: What is it used for?

src/main/java/com/easypost/model/Error.java
```diff
@@ -1,45 +1,46 @@
...
     /**
-     * Get the error code.
+     * Get the errors from an Error object.
      *
-     * @return code
+     * @return errors
      */
-    public String getCode() {
-        return code;
+    public List<Error> getError() {
```
Justintime50, line 15: Issue: This must be plural since it's a list of multiple errors

```suggestion
    public List<Error> getErrors() {
```

src/main/java/com/easypost/model/Error.java
```diff
@@ -1,45 +1,46 @@
...
-     * Set the error code.
-     *
-     * @param code code
+     * Set the errors for an Error object.
+     * 
+     * @param errors errors
      */
-    public void setCode(final String code) {
-        this.code = code;
+    public void setError(List<Error> errors) {
```
Justintime50, line 24: Issue: This must be plural since it's a list of multiple errors

```suggestion
    public void setErrors(List<Error> errors) {
```

src/main/java/com/easypost/model/Error.java
```diff
@@ -59,22 +60,4 @@ public String getMessage() {
...
-    }
-
-    /**
-     * Set the suggestion for the error.
-     *
-     * @param suggestion suggestion
-     */
-    public void setSuggestion(final String suggestion) {
-        this.suggestion = suggestion;
-    }
```
Justintime50, line 79: I believe we still need `suggestion`. It's not often used but sometimes on address verification we'll actually suggest an update to the address to make it verify correctly. (search `suggestion` in our docs and you'll find it in the responses)

src/test/java/com/easypost/ErrorTest.java
```diff
@@ -34,4 +55,93 @@ public void testError() throws EasyPostException {
         // so we'll just check fo a generic exception
         assertThrows(Exception.class, () -> Shipment.create(null));
     }
+
```
Justintime50, line 58: I can't mark the lines because they weren't edited, but now that the VCR bug is corrected, please update the `testError` test to test pulling out each individual piece of an error. You can look to the PHP or Ruby error as an example:

- https://github.com/EasyPost/easypost-ruby/blob/1673ceb6991c223ab8dc18cb1abbf2f9087e4ce1/spec/error_spec.rb

src/test/java/com/easypost/ErrorTest.java
```diff
@@ -34,4 +55,93 @@ public void testError() throws EasyPostException {
...
+    @Test
+    public void testInvalidAddressCreation() throws EasyPostException {
+        vcr.setUpTest("error_address_creation");
+        EasyPostException exception = assertThrows(EasyPostException.class,
+                () -> Address.createAndVerify(null));
+
+        assertEquals("PARAMETER.REQUIRED", exception.getCode());
+        assertEquals(422, exception.getStatusCode());
+        assertEquals("Missing required parameter.", exception.getMessage());
+    }
```
Justintime50, line 146: Issue: Please move this to the `Address` test file like other libs. It's intended to test that even though we ripped out some logic some time ago about how errors were handled for addresses that it still gets thrown correctly.

Justintime50 (changes requested):
Looks like a few things crept in. Will stamp once these are addressed.

pom.xml
jchen293: It's being used in here: https://github.com/EasyPost/easypost-java/blob/758e497082af0fe211a2919400f5c005c06d0660/src/main/java/com/easypost/utils/Cryptography.java#L3

CHANGELOG.md
jchen293: Emm, rebase gone wrong

src/main/java/com/easypost/model/Error.java
jchen293: Thanks for calling this out! Looks like I also need to bring `field` back as well

src/test/java/com/easypost/ErrorTest.java
jchen293: resolved

src/test/java/com/easypost/ErrorTest.java
jchen293: resolved

Justintime50 (approved):

src/main/java/com/easypost/exception/EasyPostException.java
```diff
@@ -8,50 +8,100 @@
 
 package com.easypost.exception;
 
+import com.easypost.model.Error;
+
+import java.util.List;
+
 public class EasyPostException extends Exception {
 
     private static final long serialVersionUID = 1L;
```
nwithan8, line 17: Reference for future travelers of what this is: https://www.baeldung.com/java-serial-version-uid

src/main/java/com/easypost/exception/General/InvalidParameterError.java
```diff
@@ -0,0 +1,24 @@
...
+        super(message);
+    }
+
+    /**
+     * constructor.
+     *
+     * @param message the exception message
+     * @param e the Throwable object
+     */
+    public InvalidParameterError(final String message, Throwable e) {
```
nwithan8, line 21: Question: Why do some of these exceptions have a Throwable constructor and others don't?
I understand what the Throwable does (basically allows for inner-nested exceptions), why are we inconsistent with which exception types can and can't have an inner Throwable?

nwithan8 (approved):

src/main/java/com/easypost/exception/General/InvalidParameterError.java
Justintime50: @jchen293 care to answer? Do we need to correct this? Pinging to make sure it doesn't fall through the cracks.

jchen293: We have a `Throwable` in this Exception class because there is a `UnsupportedEncodingException` that could possibly throw when encoding the report type, you could find the code here: https://github.com/EasyPost/easypost-java/blob/de759f2bad6f8f608591aa15a42bd0fd17ad7215/src/main/java/com/easypost/model/Report.java#L211

jchen293: @Justintime50 Had to merge first so I grab get the link to the latest code commit

PR #195, opened 2022-10-19
improve Event class

jchen293:
# Description

- Add two missing attributes in the Event class: `pendingUrls` and `completedUrls`
- Remove custom Event deserializer since the type of `result` of Event has been changed to a Map

# Testing

Current unit tests pass, add another assertion in the Webhook test that has `result` from the response to make sure it gets deserialized correctly

# Pull Request Type


Justintime50 (approved):

src/main/java/com/easypost/model/Event.java
```diff
@@ -3,14 +3,17 @@
 import com.easypost.exception.EasyPostException;
 import com.easypost.net.EasyPostResource;
 
+import java.util.List;
 import java.util.Map;
 
 public final class Event extends EasyPostResource {
     private String id;
```
nwithan8, line 10: Question: Have we not moved `id` and `mode` to the base `EasyPostResource` class yet? If so, they're not needed to be duplicated here.

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -58,11 +56,9 @@ protected enum RequestMethod {
     public static final Gson GSON =
             new GsonBuilder().setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
                     .registerTypeAdapter(HashMap.class, new HashMapSerializer())
-                    .registerTypeAdapter(Event.class, new EventDeserializer())
                     .registerTypeAdapter(SmartrateCollection.class, new SmartrateCollectionDeserializer()).create();
     public static final Gson PRETTY_PRINT_GSON = new GsonBuilder().setPrettyPrinting().serializeNulls()
-            .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
-            .registerTypeAdapter(Event.class, new EventDeserializer()).create();
+            .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES).create();
```
nwithan8, line 61: It's beautiful.

nwithan8 (approved):

src/main/java/com/easypost/model/Event.java
jchen293: We only have the getters for `id` and `mode` in `EasyPostResource`, and they are returning an empty string which is incorrect. We have to address this in a different PR.

PR #199, opened 2022-10-21
Clean up

jchen293:
# Description

DON'T review it yet

# Testing

DON'T review it yet

# Pull Request Type


src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -8,125 +8,20 @@
...
-            new ArrayList<>(Arrays.asList("getCreatedAt", "getUpdatedAt", "getFees"));
-    public static final int DEFAULT_CONNECT_TIMEOUT_MILLISECONDS = 30000;
-    public static final int DEFAULT_READ_TIMEOUT_MILLISECONDS = 60000;
-    public static final double DEFAULT_APP_ENGINE_TIMEOUT_SECONDS = 20.0;
-    private static final String DNS_CACHE_TTL_PROPERTY_NAME = "networkaddress.cache.ttl";
-    private static final String CUSTOM_URL_STREAM_HANDLER_PROPERTY_NAME = "com.easypost.net.customURLStreamHandler";
-    private static int connectTimeoutMilliseconds = DEFAULT_CONNECT_TIMEOUT_MILLISECONDS;
-    private static int readTimeoutMilliseconds = DEFAULT_READ_TIMEOUT_MILLISECONDS;
-    private static double appEngineTimeoutSeconds = DEFAULT_APP_ENGINE_TIMEOUT_SECONDS;
+public abstract class EasyPostResource extends Requestor {
```
Justintime50, line 20: I don't think this should extend Requestor. Requestor should be static HTTP logic that has nothing to do with EasyPost objects.

We may need to continue refactoring here. The problem with extending it is that all the downstream EasyPost resources will have access to whatever is available in Requestor which is not what we want.

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -795,14 +129,16 @@ private Object getIdString() {
             Field idField = this.getClass().getDeclaredField("id");
             return idField.get(this);
         } catch (SecurityException e) {
-            return "";
+            e.printStackTrace();
```
Justintime50, line 132: I'm not sure we want to do this. Printing the stack trace isn't as useful as raising an actual error here saying the object has no ID.

src/main/java/com/easypost/net/EasyPostResource.java
jchen293: We do need to extend Requestor here, or else the classes will not be able to make any HTTP requests. 

This is how it looks like for the inheritance now. The subclass(e.g Address) -> EasyPostResource(which has the shared attributes that all classes have) -> Requestor(The HTTP request functions that all classes need to make API call). Basically, we have separated our HTTP requests function to a parent class of EasyPostResource

src/main/java/com/easypost/net/EasyPostResource.java
jchen293: I think the legacy code just ignores the error and returns an empty string for `id`. To stay the same logic, we could keep the function working which we just return an empty string and print out the error stack trace

src/main/java/com/easypost/net/EasyPostResource.java
Justintime50: We do not need to include it here. It's convenient that we do but there is no requirement.

My suggestion is we do not extend Requestor here and instead import the `request` function to these resource files explicitly. It will take a bit more refactoring to import a few more things but is probably the better option.

src/main/java/com/easypost/net/EasyPostResource.java
Justintime50: We don't use `printStackTrace` anywhere else in the lib. It's meant for debugging purposes. Since this library will be used by end-users and not us, we should either leave it how it was or instead raise an actual error here with a helpful human-readable message. Sending through the whole stack trace is too messy and is unwanted from a library.

src/main/java/com/easypost/net/EasyPostResource.java
jchen293: Yes, we could do that. Actually, I should have moved `classURL()`, `instanceURL()` and enum `RequestMethod` to EasyPostResource. This will make only one import per file for the `request` function. I will address this next Monday.

nwithan8: This `Requestor` is really the origins of the `Client` that we'll eventually make anyway. No object needs to extend `Requestor`, instead each object will internally have a configured `Requestor` instance that it will use to make HTTP calls.

nwithan8 (approved):
Approving on the condition that the two comments above are addressed

src/main/java/com/easypost/net/EasyPostResource.java
jchen293: Code is reverted

src/main/java/com/easypost/net/EasyPostResource.java
jchen293: Resolved

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -17,11 +17,70 @@
...
+     * Set the timeout in milliseconds for connecting to the API.
+     *
+     * @param milliseconds the timeout in milliseconds
+     */
+    public static void setConnectTimeoutMilliseconds(int milliseconds) {
+        connectTimeoutMilliseconds = milliseconds;
+    }
+
+    /**
+     * Get the timeout in milliseconds for reading API responses.
```
Justintime50, line 68: Issue: these variables and timeout functions need to be in the Requestor class since they have nothing to do with EasyPost objects.

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -186,17 +264,22 @@ public void merge(final EasyPostResource obj, final EasyPostResource update) {
...
+                     * https://stackoverflow.com/a/7526414/13343799
+                     * Per the comments, there are some built-in expectations for how this will
+                     * work,
+                     * and should eventually be re-written or removed entirely
+                     * (explore returning a brand-new object rather than modifying the existing
+                     * one).
+                     * For now, the easiest fix would be to
+                     * a) just ignore this exception, or
+                     * b) rename getSmartrates() in the Shipment class to just smartrates()
+                     * (similar to how the other methods are named).
```
Justintime50, line 282: All these extra stars aren't necessary since a `/* ... */ will comment out everything in-between.

src/main/java/com/easypost/net/Requestor.java
```diff
@@ -48,7 +48,7 @@
 import java.util.Scanner;
 
 public abstract class Requestor {
-    protected enum RequestMethod {
+    public enum RequestMethod {
```
Justintime50, line 51: Issue: this should remain `protected`. Users will never need access to this.

src/main/java/com/easypost/net/Requestor.java
```diff
@@ -57,89 +57,10 @@ protected enum RequestMethod {
 
     private static final String DNS_CACHE_TTL_PROPERTY_NAME = "networkaddress.cache.ttl";
     private static final String CUSTOM_URL_STREAM_HANDLER_PROPERTY_NAME = "com.easypost.net.customURLStreamHandler";
-    private static int connectTimeoutMilliseconds = Constant.DEFAULT_CONNECT_TIMEOUT_MILLISECONDS;
```
Justintime50: Please put these timeout vars and functions back, they were in the correct place.

src/main/java/com/easypost/net/EasyPostResource.java
```diff
@@ -87,6 +146,27 @@ public void setUpdatedAt(final Date updatedAt) {
...
+    }
+
+    protected static String singleClassURL(final Class<?> clazz) {
+        return String.format("%s/%s", EasyPost.API_BASE, className(clazz));
+    }
+
+    private static String className(final Class<?> clazz) {
+        return clazz.getSimpleName().replaceAll("([a-z])([A-Z])", "$1_$2").toLowerCase().replace("$", "");
+    }
+
```
Justintime50, line 169: These are now in the correct place, thanks.

Justintime50 (changes requested):

src/main/java/com/easypost/net/Requestor.java
jchen293: Since we are no longer extending `Requestor` in `EasyPostResource`, we need to make this `public` so we could use this enum class in the `request` functions in every class.

src/main/java/com/easypost/net/EasyPostResource.java
jchen293: Oops, good catch. I should have updated the test file instead of moving this around.

jchen293: resolved

src/main/java/com/easypost/net/Requestor.java
jchen293: resolved

src/main/java/com/easypost/net/Requestor.java
Justintime50: I thought the whole point of protected was that anything inside our own library could use it?

jchen293: No Justin😂, `protected` means the subclass can use the property from the parent class, but since we are no longer extending to the `Requestor` class in the `EasyPostResource` file, we just lost the access permission for it. 

Link to a resource: https://www.geeksforgeeks.org/protected-keyword-in-java-with-examples/

Justintime50: Our entire library uses `package com.easypost.net;`. Based on the link you sent me, we should still be able to use `protected` here because even though we no longer inherit these, all of this functionality still lives inside the same package.

jchen293: No, we can't. They are in different packages. `Requestor` is in `com.easypost.net` while our classes are in `package com.easypost.model`.

Justintime50: Java is a terrible language.

Justintime50 (dismissed):

src/main/java/com/easypost/net/Requestor.java
nwithan8: I'd recommend using the "default" access modifier instead, so the function isn't open to end-users, just inside our library.

nwithan8 (changes requested):
Please look into making "public" functions "default" instead.

jchen293:
> 

Could you be more specific about which public functions?

jchen293:
> Please look into making "public" functions "default" instead.


Could you be more specific about which public functions?

nwithan8:
> > Please look into making "public" functions "default" instead.
> 
> Could you be more specific about which public functions?

Specifically https://github.com/EasyPost/easypost-java/pull/199#discussion_r1003585593

src/main/java/com/easypost/net/Requestor.java
jchen293: We can't use `default` since they are in different packages, please let me know which function(s) you're referring to

jchen293: I just made a small tweak, moving the `RequestMethod` enum to EasyPostResource and then importing it in the `Requestor` file since they are in the package. This will avoid end-users from using this property.

nwithan8 (approved):

Justintime50 (approved):

PR #200, opened 2022-10-27
[POC] thread-safe rewrite

jchen293:
# Description

This is a very high-level proof of concept for the `Hybrid-Based` design in the upcoming thread-safe rewrite. Please disregard the linting and code syntax at the moment.  I have duplicated the Address class for an example of how it would look without modifying the current code to avoid a dozen files being changed.

Summary of this design:
A Client instance with the API key and other configurations(timeouts) is declared. The end-user uses a resource from the client (a property), and from there calls all functions related to that resource. All API functionality is called via the resource (no instance functions). The resource uses the Client that was passed into it during initialization. We would remove all instance-based functions with this approach.

Things we need to change/add/remove in the upcoming thread-safe rewrite
1. A new `EasyPostClient` object that has `apiKey`, `connectionTimeout`, `readTimeout`. The API key parameter is always required when initializing the object, and timeout parameters are optional.
2. Split the current class (e.g Address) into `model` and `service` folders, where the `model` folder has all the getters/setters and properties of the Address class (e.g street1, street2, etc), the `service` folder will have all the functions belonging to that class. This will prevent the users to use any functions from the response object, and the `EasyPostClient` object will always be used as the root object for any HTTP request call.
3. Update the `request` functions in the `Requestor` class, the `apiKey` parameter will be updated to the `EasyPostClient object` which already has the apiKey and timeouts configuration. Remove the timeout variables and getters/setters in the `Requestor` class.
4. Update all unit tests and configure the `TestUtils` file if needed.

# Testing

# Pull Request Type

### DIFF
--- a/src/test/java/com/easypost/ThreadSafeTest.java
+++ b/src/test/java/com/easypost/ThreadSafeTest.java
@@ -1,0 +1,55 @@
+package com.easypost;
+
+import com.easypost.exception.EasyPostException;
+import com.easypost.model.Address;
+import com.easypost.model.AddressThreadSafe;
+import com.easypost.service.EasyPostClient;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.BeforeAll;
+
+public class ThreadSafeTest {
+    private static TestUtils.VCR vcr;
+
+    @BeforeAll
+    public static void setup() throws EasyPostException {
+        vcr = new TestUtils.VCR("address_thread_safe", TestUtils.ApiKey.TEST);
+    }
+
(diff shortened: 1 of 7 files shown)


src/test/java/com/easypost/ThreadSafeTest.java
```diff
@@ -0,0 +1,55 @@
...
+        vcr = new TestUtils.VCR("address_thread_safe", TestUtils.ApiKey.TEST);
+    }
+
+    // Test creating/retrieving/verifying an address (thread-safe) with EasyPostClient object
+    @Test
+    public void testAddress() throws EasyPostException {
+        vcr.setUpTest("create");
+        EasyPostClient client = new EasyPostClient(System.getenv("EASYPOST_TEST_API_KEY"));
+        
+        AddressThreadSafe address = client.Address().create(Fixtures.caAddress1());
```
Justintime50, line 31: Thought: I don't think we'd want to have user instantiating each namespace, we should be doing this behind the scenes for them when it's called right? Seems like a lot of overhead that could easily be forgotten.

jchen293: Not sure if I follow. Could you provide an example?

PR #204, opened 2022-11-08
Add thread-safe to the library

jchen293:
# Description

This PR adds thread-safe to the EasyPost Java client library, as well as some other minor improvements

- A new EasyPostClient object that has apiKey, connectionTimeout, and readTimeout. The API key parameter is always required when initializing the object, and timeout parameters are optional.
- Split the current class (e.g Address) into model and service folders, where the model folder has all the getters/setters and properties of the Address class (e.g street1, street2, etc), the service folder will have all the functions belonging to that class. This will prevent the users to use any functions from the response object, and the EasyPostClient object will always be used as the root object for any HTTP request call.
- Update the request functions in the Requestor class, the apiKey parameter will be updated to the EasyPostClient object which already has the apiKey and timeouts configuration. Remove the timeout variables and getters/setters in the Requestor class.
- Update all unit tests and configure the TestUtils file.
- Remove the `merge` function and associate helper functions because it's no longer needed as we decided to return the response value directly instead of updating the object.
- AppEngine timeout is now set to the connection timeout.
- Remove an invalid function `User.create()` since creating a child-user always requires a name.
- Move helper functions (instanceUrl, classUrl) to the `Utilities` file from `EasyPostResource` because model classes no longer use it.

# Testing

Re-recorded current unit tests
Add a unit test for EasyPostClient

# Pull Request Type


nwithan8 (changes requested):
Please update the examples in the README

nwithan8:
Have we accounted for beta and API versioning in this? We can kick that down the road if needed.

jchen293:
> Have we accounted for beta and API versioning in this? We can kick that down the road if needed.

It's not part of thread-safe PR because I would like to keep the PR focusing on thread-safe only. Once we have another Beta endpoint available for us, we will make a separate PR

nwithan8 (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 ## V6 (Next release)
 
+- Library is now thread-safe
+  - Initialize a `EasyPostClient` object with an API key, connection and readtime milliseconds are optional, otherwise it will be set to default.
+  - All methods (i.e. `create`, `retrieve`, retrieve `all` of a resource) exist in services, accessed via property of the client
+    - E.g. Static method -> `client.address.create(params)`
+    - E.g. Instance method -> `client.shipment.buy(lowestRate, shipmentID)`
+  - All functions are now taking object ID instead of an object, which means the functions are returning the response value directly instead of updating the current object
+    - E.g. `client.shipment.buy(lowestRate, shipmentID)`
```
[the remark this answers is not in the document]
jchen293: resolved, good catch!

nwithan8 (approved):

src/main/java-templates/com/easypost/EasyPost.java
```diff
@@ -33,8 +33,6 @@ public abstract class EasyPost {
     public static final String VERSION = "${project.version}";
     public static final String API_BASE = "https://api.easypost.com/v2";
     public static final String BETA_API_BASE = "https://api.easypost.com/beta";
```
Justintime50, line 35: Question: Do we want to move these API base strings elsewhere? The client object or the Constant file instead?

src/main/java/com/easypost/http/Requestor.java
```diff
@@ -428,18 +448,25 @@ protected static <T> T request(final RequestMethod method, final String url,
...
-            }
-        }
-
-        if (apiKey == null) {
-            apiKey = EasyPost.apiKey;
+    protected static <T> T httpRequest(final RequestMethod method, final String url,
+            final Map<String, Object> params, final Class<T> clazz, final EasyPostClient client)
+            throws EasyPostException {
+        if (client.getApiKey() == null || client.getApiKey().isEmpty()) {
+            throw new MissingParameterError(Constants.INVALID_API_KEY_TYPE);
```
Justintime50, line 469: This check should happen higher up the stack - you can't create a client object without an API key, so this check shouldn't be necessary and can be removed (it's a duplicate).

src/main/java/com/easypost/model/CreditCard.java
```diff
@@ -1,14 +1,5 @@
 package com.easypost.model;
```
Justintime50, line 1: Let's completely remove the `CreditCard` class and functionality. I know it will be a "version early", but it was added by accident and should be super low impact since it was a one-off. Let's just call it out in the CHANGELOG please.

src/main/java/com/easypost/model/PaymentMethodObject.java
```diff
@@ -172,7 +172,7 @@ public PaymentMethodType getType() {
      *
      * @return the endpoint for this PaymentMethodObject.
      */
-    String getEndpoint() {
+    public String getEndpoint() {
```
Justintime50, line 175: I hate that Java doesn't have a scope where you can use a function in your own package but block it from outsiders just because it's in a different folder...

src/main/java/com/easypost/service/CreditCardService.java
```diff
@@ -0,0 +1,85 @@
+package com.easypost.service;
```
Justintime50, line 1: As mentioned on the model, let's remove the `CreditCard` class and logic completely and call it out in the CHANGELOG.

src/main/java/com/easypost/service/EasyPostClient.java
```diff
@@ -0,0 +1,118 @@
+package com.easypost.service;
+
+import com.easypost.http.Constant;
+
+public class EasyPostClient {
+    public final AddressService address;
+    public final ApiKeyService apikeys;
+    public final UserService user;
```
Justintime50, line 8: Sort this list please 😉

src/test/java/com/easypost/BatchTest.java
```diff
@@ -182,21 +182,21 @@ public void testCreateScanForm() throws EasyPostException, InterruptedException
...
         shipmentData.add(shipment);
         params.put("shipments", shipmentData);
 
-        Batch batchWithAddedShipment = batch.addShipments(params);
+        Batch batchWithAddedShipment = vcr.client.batch.addShipments(params, batch.getId());
 
         assertEquals(1, batchWithAddedShipment.getNumShipments().intValue());
 
-        Batch batchWithoutShipment = batch.removeShipments(params);
+        Batch batchWithoutShipment = vcr.client.batch.removeShipments(params, batch.getId());
```
Justintime50, line 199: Hmm, the ID of an object should always be the first parameter. Param ordering should be tied to the flow. We'll always need to know the object in question (Batch) and then we do stuff to it (add/remove shipments). Other params may come and go but if we are going this route, the ID will always be required.

This is true of all the previous instance based functions. You'll want to check them all.

src/test/java/com/easypost/ThreadTest.java
```diff
@@ -2,6 +2,8 @@
 
 import com.easypost.exception.EasyPostException;
 import com.easypost.model.Order;
+import com.easypost.service.EasyPostClient;
```
Justintime50, line 5: I'm not seeing a thread test that spins up multiple threads with various API keys and gets the API keys out of those client objects to ensure they didn't get mixed up, can we please add that here?

Justintime50 (commented):
I didn't check the rest of the tests (checked a couple), will do that once the rest of this feedback is addressed.

src/main/java-templates/com/easypost/EasyPost.java
nwithan8: Jack and I discussed this yesterday. I know in C# we wanted to provide a way for users to override the base URL. Do we still want that here as well?

jchen293: This is currently being used in the Utilities file `singleClassURL` function. Should we allow the user to override the base URL?

src/main/java/com/easypost/http/Requestor.java
jchen293: Users could initialize a EasyPostClient object with a `null` or `empty` string, this check statement is needed in case a user pass in an invalid `API Key` when creating EasyPostClient

nwithan8: The null check should happen (and throw an error) during Client construction.

jchen293: Yea, we could. But the user will need to add an exception handler in the EasyPostClient initialization method. Any thought?

src/main/java-templates/com/easypost/EasyPost.java
Justintime50: Oh yeah, yes please. It should be an optional param to the Client object. Remove the `/v2` from the base URL since that's the version and make sure the version is a private prop of the Client so only we can set it

src/main/java/com/easypost/http/Requestor.java
Justintime50: API key is required, so don't let users create a Client without one (don't provide that overload)

nwithan8: If it throws an error because the user did something incorrectly, it's on the user to account for doing things incorrectly, you know?

src/test/java/com/easypost/BatchTest.java
jchen293: Not sure if I follow, but what is the benefit of swapping `param` and `id` as both parameters are required?

src/test/java/com/easypost/ThreadTest.java
jchen293: This is the old unit test, there is a test to ensure API keys do not get mixed up between multiple client objects in `EasyPostTest.java`.

src/test/java/com/easypost/BatchTest.java
Justintime50: Consistency across all functions like this. ID will always be required for these previous instance methods whereas some of these will have optional params. If ID comes first, then it's always the same order for all functions.

src/test/java/com/easypost/ThreadTest.java
Justintime50: If so, the thread test should be placed in the thread test test file, don't you agree?

src/main/java-templates/com/easypost/EasyPost.java
jchen293: This will be the fourth parameter in the EasyPostClient constructor, are we allowing users to override the default `/v2` version?

Justintime50: No, I said:

> make sure the version is a private prop of the Client so only we can set it

jchen293: Ok, no optional param should be given in the constructor.

src/test/java/com/easypost/ThreadTest.java
jchen293: sure

src/test/java/com/easypost/BatchTest.java
jchen293: Just looked at Go, indeed we the ID comes first before any other parameters

nwithan8:
`apiBase` should be, e.g., "https://api.easypost.com", and should be customizable in the Client constructor.

`apiVersion` should be, e.g., "v2" and is set internally (end-user doesn't know about it).

jchen293:
> `apiBase` should be, e.g., "https://api.easypost.com", and should be customizable in the Client constructor.
> 
> `apiVersion` should be, e.g., "v2" and is set internally (end-user doesn't know about it).

If `apiBase` is customizable, what if the user overrides with the wrong API base URL?

nwithan8:
> > `apiBase` should be, e.g., "https://api.easypost.com", and should be customizable in the Client constructor.
> > 
> > `apiVersion` should be, e.g., "v2" and is set internally (end-user doesn't know about it).
> 
> If `apiBase` is customizable, what if the user overrides with the wrong API base URL in the Client constructor?
> 

Then their requests will probably fail. Again, it's on the end user to account for that if they are going to utilize "advanced features" like that.

jchen293:
> > > `apiBase` should be, e.g., "https://api.easypost.com", and should be customizable in the Client constructor.
> > > `apiVersion` should be, e.g., "v2" and is set internally (end-user doesn't know about it).
> > 
> > 
> > If `apiBase` is customizable, what if the user overrides with the wrong API base URL in the Client constructor?
> 
> Then their requests will probably fail. Again, it's on the end user to account for that if they are going to utilize "advanced features" like that.

I am not a big fan of this. We probably don't have much `API base`, I think we could use Enum for the `apiBase` optional parameter, this would prevent spelling mistakes like typos from the user input.

nwithan8:
> > > > `apiBase` should be, e.g., "https://api.easypost.com", and should be customizable in the Client constructor.
> > > > `apiVersion` should be, e.g., "v2" and is set internally (end-user doesn't know about it).
> > > 
> > > 
> > > If `apiBase` is customizable, what if the user overrides with the wrong API base URL in the Client constructor?
> > 
> > Then their requests will probably fail. Again, it's on the end user to account for that if they are going to utilize "advanced features" like that.
> 
> I am not a big fan of this. We probably don't have much `API base`, I think we could use Enum for the `apiBase` optional parameter, this would prevent spelling mistakes like typos from the user input.

See how we do it in C#: https://github.com/EasyPost/easypost-csharp/blob/master/EasyPost/Client.cs#L63

Justintime50:
@jchen293 we want to allow users to override the base API for two reasons:

1. We could point the client library at our internal test servers if needed for end to end test something
2. If we ever released a mock API product in the future, users could override the base API to hit that instead (for testing or setup purposes)

It should be a free-hand string users can specify. It will not be documented as the use cases will be slim; however, some of our other libs like Python allow you to override this already and we'd like them all to allow it.

As for the versioning, client libs will always be locked to the most recent version available and this cannot be customized. In theory, whatever base APIs are available will have the same set of API versions available to them.

jchen293:
> @jchen293 we want to allow users to override the base API for two reasons:
> 
> 1. We could point the client library at our internal test servers if needed for end to end test something
> 2. If we ever released a mock API product in the future, users could override the base API to hit that instead (for testing or setup purposes)
> 
> It should be a free-hand string users can specify. It will not be documented as the use cases will be slim; however, some of our other libs like Python allow you to override this already and we'd like them all to allow it.
> 
> As for the versioning, client libs will always be locked to the most recent version available and this cannot be customized. In theory, whatever base APIs are available will have the same set of API versions available to them.

Got it, we probably don't want to expose the `API Base` in the source code by using Enum, I will add another optional parameter for the API base next week.

src/main/java/com/easypost/service/ReferralCustomerService.java
```diff
@@ -220,7 +222,8 @@ private PaymentMethodObject createEasypostCreditCard(final String referralApiKey
         wrappedParams.put("credit_card", params);
 
         EasyPostClient referralClient = new EasyPostClient(referralApiKey);
-        return Requestor.request(RequestMethod.POST, String.format("%s/%s", EasyPost.API_BASE, "credit_cards"),
+        return Requestor.request(RequestMethod.POST,
+                String.format("%s/%s/%s", client.getApiBase(), client.getApiVersion(), "credit_cards"),
```
Justintime50, line 226: I'd prefer we set all these `String.format` items to an intermediary variable so we know what it is and these lines can be much shorter. Thoughts?

If you agree, we should do that throughout the whole project. It just seems crammed with so much text on one line which then wraps to two or three..

jchen293: I agree, this is resolved.

Justintime50 (approved):
I thiiink all of my feedback was addressed. This is good work!

src/main/java/com/easypost/service/EasyPostClient.java
```diff
@@ -125,8 +166,17 @@ public String getApiKey() {
     }
 
     /**
-     * Get API base for this EasyPostClient object.
+     * Get API version for this EasyPostClient object.
      * 
+     * @return the API version for this EasyPostClient object.
+     */
+    public String getApiVersion() {
```
nwithan8, line 173: We don't need getters for the API version. Make the `apiVersion` property public and final, so it can't be overridden by end-users (only set via the constructor)

src/main/java/com/easypost/service/EasyPostClient.java
```diff
@@ -7,8 +7,9 @@
 public class EasyPostClient {
     private final int connectTimeoutMilliseconds;
     private final int readTimeoutMilliseconds;
-    private final String apiBase = "v2";
     private final String apiKey;
+    private final String apiVersion = "v2";
```
nwithan8, line 11: Make `apiVersion` public (see below)

src/main/java/com/easypost/service/EasyPostClient.java
```diff
@@ -7,8 +7,9 @@
...
+    private String apiBase;
```
nwithan8, line 12: Make this final, since it won't change after being set by the constructor.

src/main/java/com/easypost/service/BillingService.java
```diff
@@ -65,20 +65,22 @@ public void fundWallet(String amount, PaymentMethod.Priority priority)
...
      *
      * @return an EasyPost.PaymentMethod summary object.
-     * @throws EasyPostException when the request fails or billing has not been set up.
+     * @throws EasyPostException when the request fails or billing has not been set
+     *                           up.
      */
     public PaymentMethod retrievePaymentMethods() throws EasyPostException {
         PaymentMethod response = Requestor.request(RequestMethod.GET,
-                String.format("%s/%s", EasyPost.API_BASE, "payment_methods"), null,
+                String.format("%s/%s/%s", client.getApiBase(), client.getApiVersion(), "payment_methods"), null,
```
nwithan8: Since the API base won't change for different API calls (it's not like one API call is going to go to a different base url than every other API call + the fact that API base can't be changed after construction anyway), we can move API base formatting downstream into the `request` function.

API version can change on a per-call basis (some go to Beta, some go to v2, etc), that should remain here as a parameter.

src/main/java/com/easypost/service/EasyPostClient.java
```diff
@@ -125,8 +166,17 @@ public String getApiKey() {
...
+    public String getApiVersion() {
+        return apiVersion;
+    }
+
+    /**
+     * Get API base for this EasyPostClient object.
+     *
      * @return the API base for this EasyPostClient object.
      */
     public String getApiBase() {
```
nwithan8, line 182: We don't need getters for the API base. Make the `apiBase` property public and final, so it can't be overridden by end-users (only set via the constructor)

src/main/java/com/easypost/service/EasyPostClient.java
nwithan8: Alternatively, if we are going to keep this getter, let's have it accept a parameter so it can return `v2` by default, but other possible versions (i.e. `beta`) on request.

src/main/java/com/easypost/service/EasyPostClient.java
```diff
@@ -44,30 +45,70 @@ public EasyPostClient(String apiKey) throws MissingParameterError {
...
+    /**
+     * EasyPostClient constructor.
+     *
+     * @param apiKey                     API key for API calls.
+     * @param connectTimeoutMilliseconds Timeout for connection.
+     * @param readTimeoutMilliseconds    Timeout for read.
+     * @param apiBase                    API base for API calls.
+     * @throws MissingParameterError
+     */
+    public EasyPostClient(String apiKey, int connectTimeoutMilliseconds, int readTimeoutMilliseconds, String apiBase)
```
nwithan8, line 105: Nit: I hate all these overloads. Have an idea for a better way, but won't block on it.

nwithan8 (changes requested):

src/main/java/com/easypost/service/EasyPostClient.java
jchen293: I think this is just our personal preference, I like to use getter/setter for immutable variables in Java to follow the `encapsulation` practice, unless there is a significant advantage by making these two fields `public final` then I am down to make that change.

src/main/java/com/easypost/service/EasyPostClient.java
jchen293: see below answer

src/main/java/com/easypost/service/EasyPostClient.java
jchen293: We should not have an additional parameter in the getter method, this will allow end users to change the apiVersion variable. The reason why the `apiVersion` is not final so we could change it internally in the `EasyPostClient` if the namespace has `beta`, e.g: `client.beta.endshipper.create()`

src/main/java/com/easypost/service/EasyPostClient.java
nwithan8: Fair enough, won't block.

src/main/java/com/easypost/service/BillingService.java
jchen293: Requestor class has it, in line 429. This function is not using the `classURL()` from the `Utilities` file because the endpoint has a different name from the model class name (Billing), that's why we are passing in the `API base` here in the `String.format`

src/main/java/com/easypost/service/EasyPostClient.java
jchen293: see above answer

src/main/java/com/easypost/service/EasyPostClient.java
jchen293: So am I? It's Java things

src/main/java/com/easypost/service/EasyPostClient.java
nwithan8: At least make the apiBase final

src/main/java/com/easypost/service/BillingService.java
nwithan8: Not a fan of the classURL() utility if it's going to cause confusing edge cases like this when we DON'T use it. Won't block, but could do with a re-imagining down the line.

src/main/java/com/easypost/service/EasyPostClient.java
nwithan8: I foresee some issues arising when we do need to implement a new beta service down the road with this current design, but we'll kick the can on that for now.

nwithan8 (approved):

src/main/java/com/easypost/service/EasyPostClient.java
jchen293: I will make this tiny change in next PR so I don't dismiss the approvals

PR #207, opened 2022-11-22
robust test coverage

jchen293:
# Description

Increase code coverage from ~45% to 90%, this excludes all the getters/setters and the `makeAppEngine` function in the Requestor. 
Use the Mockito framework for mocking.

Squash a few bugs
- Removes the `refresh()` function in all classes since every method is now static, please use `retrieve()` function instead
- Removes invalid functions that allow users to make an API request without providing the required parameters
- Move the lowest_rate functions from service to model for `Order` and `PickUp` since no API interaction is involved

Screen Shot 2022-11-23 at 5 23 19 PM

# Testing

Run the test coverage with Jacoco and the code coverage is now about 90%

# Pull Request Type


src/main/java/com/easypost/http/Requestor.java
```diff
@@ -81,10 +83,6 @@ static Map<String, String> generateHeaders(String apiKey) throws MissingParamete
                 convertSpaceToHyphen(System.getProperty("os.arch")),
                 convertSpaceToHyphen(System.getProperties().getProperty("java.vm.name"))));
 
-        if (apiKey == null || apiKey.isEmpty()) {
```
jchen293, line 84: # Reviewer note:

This is no longer needed since there is a check statement in the constructor when initializing the `EasyPostClient` object

src/test/java/com/easypost/BillingTest.java
```diff
@@ -22,17 +36,38 @@ public static void setup() throws EasyPostException {
         vcr = new TestUtils.VCR("billing", TestUtils.ApiKey.PRODUCTION);
     }
 
+    /**
+     * Release the static mock once it has been used.
+     */
+    @AfterAll
+    public static void cleanup() {
+        requestMock.close();
+    }
```
Justintime50, line 45: This function should be at the top of this file so it's easy to find and so it's not mixed in with the rest of the tests.

src/test/java/com/easypost/PaymentMethodTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Test retrieving all payment methods.
+     *
+     * @throws EasyPostException when the request fails.
+     */
+    @Test
+    public void testAll() throws EasyPostException {
+        vcr.setUpTest("all");
+        // Test deprecated function all() in PaymentMethodService
```
Justintime50, line 33: This comment shouldn't be needed

src/test/java/com/easypost/ShipmentTest.java
```diff
@@ -403,6 +426,62 @@ public void testInstanceLowestSmartRate() throws EasyPostException {
...
+
+    /**
+     * Test getting smart rates for a shipment.
+     * 
+     * @throws EasyPostException
+     */
+    @Test
+    public void testGetSmartRate() throws EasyPostException {
+        vcr.setUpTest("get_smartrate_list");
+        // Test deprecated getSmartrates() function
```
Justintime50, line 454: These comments shouldn't be needed since the test name/logic would already explain this. Once the deprecated function gets removed, we'd simply remove these tests because they'd start failing once the logic gets ripped out.

TLDR: It's more clutter than provides value

Justintime50 (commented):
A few small items, otherwise looks good!

src/test/java/com/easypost/BillingTest.java
jchen293: This function is on the top of this file, it should come after `setup()` which has `BeforeAll` annotation. 
`AfterAll` function should come after the `BeforeAll` function.

src/test/java/com/easypost/PaymentMethodTest.java
jchen293: removed

src/test/java/com/easypost/ShipmentTest.java
jchen293: Removed

src/test/java/com/easypost/BillingTest.java
Justintime50: Ah, showed up weird on GitHub. Thanks!

Justintime50 (approved):

src/main/java/com/easypost/http/Requestor.java
```diff
@@ -566,6 +564,7 @@ protected static void handleAPIError(String rBody, final int rCode) throws EasyP
         }
     }
 
+    @Generated // Exclude from the jacoco test coverage
```
nwithan8, line 567: Nit: I would imagine there's a more appropriate annotation to exclude a code section from unit test coverage, instead of piggybacking off the Generated annotation. Won't block.

src/main/java/com/easypost/model/Order.java
```diff
@@ -20,4 +23,38 @@ public final class Order extends EasyPostResource {
...
+     * Get the lowest rate for this Order.
+     *
+     * @param carriers The carriers to use in the filter.
+     * @param services The services to use in the filter.
+     * @return Lowest Rate object
+     * @throws EasyPostException when the request fails.
+     */
+    public Rate lowestRate(final List<String> carriers, final List<String> services)
+            throws EasyPostException {
+        return Utilities.getLowestObjectRate(this.getRates(), carriers, services);
```
nwithan8, line 47: Can we put the base function above the two overloads for this?

src/main/java/com/easypost/model/Pickup.java
```diff
@@ -17,4 +22,44 @@ public final class Pickup extends EasyPostResource {
...
+
+    /**
+     * Get the lowest rate for this Pickup.
+     *
+     * @param carriers The carriers to use in the filter.
+     * @param services The services to use in the filter.
+     * @return lowest PickupRate object
+     * @throws EasyPostException when the request fails.
+     */
+    public PickupRate lowestRate(final List<String> carriers, final List<String> services)
```
nwithan8, line 44: Same about base-overload order here

nwithan8 (approved):

src/main/java/com/easypost/http/Requestor.java
jchen293: Jacoco released this under 0.8.2 new features: https://github.com/jacoco/jacoco/releases/tag/v0.8.2

src/main/java/com/easypost/model/Order.java
jchen293: Sure, I will adjust this in the next PR so I don't dismiss approvals.

src/main/java/com/easypost/model/Pickup.java
jchen293: Sure, I will adjust this in the next PR so I don't dismiss approvals.

PR #208, opened 2022-11-29
Add content to upgrade guide for v6 major bump

jchen293:
# Description

Add content for the V6 major bump in the upgrade guide
Move the `ExternalApiError` from the general to API folder

# Testing

N/A

# Pull Request Type

### DIFF
--- a/UPGRADE_GUIDE.md
+++ b/UPGRADE_GUIDE.md
@@ -116,9 +116,9 @@ Static and instance-based methods have been divided into separate services and r
 
 For example:
 
-- The static method `Shipment.buy()` is now accessible in the Shipment service via `client.shipment.buy("shp_...")`. 
+- The static method `Shipment.buy()` is now accessible in the Shipment service via `client.shipment.buy("shp_...")`.
 - The instance method `myShipment.lowestRate()` is still accessible only via a Shipment instance because there is no API interaction.
-- The instance method of `refresh()` has been removed becasue you can no longer use object to make a HTTP request.
+- The instance method of `refresh()` has been removed because you can no longer use an object to make a HTTP request.
 
 Instances of an object cannot be initialized directly. Instead, use the service to create the instance via a `create`, `retrieve`, or `all` API call.
 
@@ -126,13 +126,13 @@ Instances of an object cannot be initialized directly. Instead, use the service 
 
 *Likelihood of Impact: **Medium***
 
-Each object no longer has setters. The use case of the setter is extremely low, but if you need to use the setter for a specific reason, try to use `reflection` in Java.
+Objects no longer have setters. The use case of the setter is extremely low, but if you need to use the setter for a specific reason, try to use `reflection` in Java.
 
 ## 6.0 Rename Some Getters
 
 *Likelihood of Impact: **Medium***
 
-Below getters have been renamed:
+The following getters have been renamed:
 
 - Pickup class: `getPickoutRates()` -> `getPickupRates()`
 - PaymentMethod class: `getPrimaryPaymentMethodObject()` -> `getPrimaryPaymentMethod()`
@@ -142,7 +142,7 @@ Below getters have been renamed:
 
 *Likelihood of Impact: **Low***
 
-Change the below type of objects:
+The following properties have changed types:
 
 - `result` of Event from `EasyPostResource` to `Map<String, object>`
 - `amount` of Insurance from `Float` to `String`
--- a/UPGRADE_GUIDE.md
+++ b/UPGRADE_GUIDE.md
@@ -2,22 +2,130 @@
 
 Use the following guide to assist in the upgrade process of the `easypost-java` library between major versions.
 
-* [Upgrading from 4.x to 5.0](#upgrading-from-4x-to-50)
+- [Upgrading from 5.x to 6.0](#upgrading-from-5x-to-60)
+- [Upgrading from 4.x to 5.0](#upgrading-from-4x-to-50)
+
+## Upgrading from 5.x to 6.0
+
+### 6.0 High Impact Changes
+
+- [Client Instance](#60-client-instance)
+- [Error Types](#60-error-types)
+- [Uniform empty response](#60-uniform-empty-response)
+
+### 6.0 Low Impact Changes
+
+- [Services and Resources](#60-services-and-resources)
+
+## 6.0 Client Instance
+
+*Likelihood of Impact: **High***
+
+The library is now designed around the idea of a `EasyPostClient`. Users will initialize a `EasyPostClient` instance with their API key
+and then use that instance to make API calls.
+
+This change allows for multiple clients to be instantiated with different API keys and for the library to be used in a
+multi-threaded environment.
+
+```java
+// Old
+EasyPost.apiKey = System.getenv("EASYPOST_API_KEY"); // global API key
+
+// New
+EasyPostClient client1 = new EasyPostClient("EASYPOST_API_KEY_1"); // per-client API key
+EasyPostClient client2 = new EasyPostClient("EASYPOST_API_KEY_2"); // per-client API key
+```
+
+All functions are now accessed via the `EasyPostClient` instance. For example, to create a shipment:
+
+```java
+// Old
+EasyPost.apiKey = System.getenv("EASYPOST_API_KEY");
+Shipment shipment = Shipment.create(shipmentMap);
+
+// New
+EasyPostClient client = new EasyPostClient("EASYPOST_API_KEY_1");
+Shipment shipment = client.shipment.create(shipmentMap);
+```
+
+## 6.0 Error Types
+
+*Likelihood of Impact: **High***
+
+Error handling has been overhauled and a number of specific exception types have been introduced.
+
+All exceptions inherit from `EasyPost.Exception.EasyPostException` (which extends `System.Exception`). Users can catch this exception type to handle all errors thrown by the library.
+
+Subclasses of `EasyPostException` are grouped into two categories:
+
+- `EasyPost.Exception.API` for errors returned by the API. Subclasses of this exception type are:
+  - `EasyPost.Exception.ExternalApiError` - thrown when an issue occurs with an external API (e.g. Stripe)
+  - `EasyPost.Exception.ForbiddenError` - thrown when you don't have permission to access this API resource
+  - `EasyPost.Exception.GatewayTimeoutError` - thrown when the API gateway times out (504 status code)
+  - `EasyPost.Exception.InternalServerError` - thrown when an internal server error occurs (500 status code)
+  - `EasyPost.Exception.InvalidRequestError` - thrown when the API received an invalid request (422 status code)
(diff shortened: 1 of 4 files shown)


UPGRADE_GUIDE.md
```diff
@@ -2,22 +2,130 @@
...
+- [Upgrading from 5.x to 6.0](#upgrading-from-5x-to-60)
+- [Upgrading from 4.x to 5.0](#upgrading-from-4x-to-50)
+
+## Upgrading from 5.x to 6.0
+
+### 6.0 High Impact Changes
+
+- [Client Instance](#60-client-instance)
+- [Error Types](#60-error-types)
+- [Uniform empty response](#60-uniform-empty-response)
```
Justintime50, line 14: We know it as uniform empty responses, but let's say what it is from the users perspective. Please fix the links too

```suggestion
- [Empty Response Function Return Types](#60-uniform-empty-response)
```

UPGRADE_GUIDE.md
```diff
@@ -2,22 +2,130 @@
...
 
-* [Upgrading from 4.x to 5.0](#upgrading-from-4x-to-50)
+- [Upgrading from 5.x to 6.0](#upgrading-from-5x-to-60)
+- [Upgrading from 4.x to 5.0](#upgrading-from-4x-to-50)
+
+## Upgrading from 5.x to 6.0
+
+### 6.0 High Impact Changes
+
+- [Client Instance](#60-client-instance)
```
Justintime50, line 12: Question: Does `Client Object` make more sense here? I'm not married to either but a thought?

UPGRADE_GUIDE.md
```diff
@@ -2,22 +2,130 @@
...
+
+### 6.0 Low Impact Changes
+
+- [Services and Resources](#60-services-and-resources)
+
+## 6.0 Client Instance
+
+*Likelihood of Impact: **High***
+
+The library is now designed around the idea of a `EasyPostClient`. Users will initialize a `EasyPostClient` instance with their API key
```
Justintime50, line 24:
```suggestion
The library is now designed around the idea of an `EasyPostClient`. Users will initialize an `EasyPostClient` instance with their API key
```

UPGRADE_GUIDE.md
```diff
@@ -2,22 +2,130 @@
...
+and then use that instance to make API calls.
+
+This change allows for multiple clients to be instantiated with different API keys and for the library to be used in a
+multi-threaded environment.
```
Justintime50, line 28:
```suggestion
This change allows for multiple clients to be instantiated with different API keys which allows this library to be safely used in a
multi-threaded environment.
```

UPGRADE_GUIDE.md
```diff
@@ -2,22 +2,130 @@
...
+
+## 6.0 Error Types
+
+*Likelihood of Impact: **High***
+
+Error handling has been overhauled and a number of specific exception types have been introduced.
+
+All exceptions inherit from `EasyPost.Exception.EasyPostException` (which extends `System.Exception`). Users can catch this exception type to handle all errors thrown by the library.
+
+Subclasses of `EasyPostException` are grouped into two categories:
```
Justintime50, line 59: - Was all of this copied from the C# guide? 
- If so, are we sure we included every error type it had? 
- Did we double check that the syntax is correct for Java?

UPGRADE_GUIDE.md
```diff
@@ -2,22 +2,130 @@
...
+- `String Message` - the error message returned by the API (e.g. "PARAMETER.INVALID")
+- `List<Error> Errors` - a list of errors returned by the API (e.g. "Invalid parameter: to_address")
+
+Users can better anticipate exception information through utilities in the `EasyPost.Exception.Constants` file.
+
+## 6.0 Uniform empty response
+
+*Likelihood of Impact: **High***
+
+Below functions return empty bodies or boolean have been changed to void.
```
Justintime50, line 101:
```suggestion
The following functions return an empty body from the API and have been changed to return void to reflect this.
```

UPGRADE_GUIDE.md
```diff
@@ -2,22 +2,130 @@
...
+
+- `fundWallet()` and `deletePaymentMethod()` in Billing class
+- `createList()` in Tracker class
+- `updateEmail()` in ReferralCustomer class
+
+## 6.0 Services and Resources
+
+*Likelihood of Impact: **Low***
+
+Static and instance-based methods have been divided into separate services and resources, methods that have API interactions are in services now. For example, the static method `Shipment.Create` is now accessible in the Shipment service via `client.Shipment.Create`. The instance method `myShipment.lowestRate` is still accessible only via a Shipment instance.
```
Justintime50, line 111: Shouldn't this be a lowercase `S`?

```suggestion
Static and instance-based methods have been divided into separate services and resources, methods that have API interactions are in services now. For example, the static method `Shipment.Create` is now accessible in the Shipment service via `client.shipment.Create`. The instance method `myShipment.lowestRate` is still accessible only via a Shipment instance.
```

UPGRADE_GUIDE.md
```diff
@@ -2,22 +2,130 @@
...
+
+Instances of an object cannot be initialized directly. Instead, use the service to create the instance via a `Create`, `Retrieve`, or `All` API call.
```
Justintime50, line 113: I'd argue that the likelyhood of this change is `high` because there are many function calls that are no longer instance-based. Please bump this up in the list and provide an example of something that used to be instance but is now static. eg: buying a shipment is no longer `shipment.buy()` but is now `client.shipment.buy("shp_...")`

UPGRADE_GUIDE.md
```diff
@@ -2,22 +2,130 @@
...
+        404 status code)
+  - `EasyPost.Exception.PaymentError` - thrown when a payment error occurs (402 status code)
+  - `EasyPost.Exception.ProxyError` - thrown when the library is unable to connect to the API via a proxy
+  - `EasyPost.Exception.RateLimitError` - thrown when the API rate limit is exceeded (429 status code)
+  - `EasyPost.Exception.RedirectError` - thrown when the http status is between 300 and 308.
+  - `EasyPost.Exception.ServiceUnavailableError` - thrown when the API is unavailable (503 status code)
+  - `EasyPost.Exception.TimeoutError` - thrown when the API request times out (408 status code)
+  - `EasyPost.Exception.UnauthorizedError` - thrown when the API receives an unauthorized request (401 or 403 status
+        code)
+  - `EasyPost.Exception.UnknownApiError` - thrown when an unknown API error occurs (unexpected 4xx status code)
```
nwithan8, line 79: Is the name space `Exception` or `Exceptions`? Inconsistent here, double-check throughout.

Justintime50 (changes requested):
I'd argue we need the following items in the upgrade guide too:

- Changes the type of Insurance `Amount` from Float to String
- Removes all the setters of each object (we say setters are available via lombok; however, I don't think we added the decorator meaning users can't access them. In the off chance someone was previously, we need to call it out that they can't any longer)
- Rename some getters
  - Pickup class: `getPickoutRates()` -> `getPickupRates()`
  - PaymentMethod class: `getPrimaryPaymentMethodObject()` -> `getPrimaryPaymentMethod()`
  - PaymentMethod class: `getSecondaryPaymentMethodObject()` -> `getSecondaryPaymentMethod()`

The point of the upgrade guide is to callout anything that will require a user to change their code and provide guidance and examples on how to do it to ease the process of moving from one major version to another. We don't need to call out additions here (the CHANGELOG can handle that), but the upgrade guide should include anything substantial that will require an engineer to make a tweak in a codebase downstream from us.

nwithan8 (changes requested):

UPGRADE_GUIDE.md
jchen293: Exception in Java

UPGRADE_GUIDE.md
jchen293: I was following the C# upgrade guide to be consistent. What if we name the title to `Thread-safe`, I think this is more sense than `Client Instance/ Object`?

UPGRADE_GUIDE.md
jchen293: Yes, it was copied from the C# guide but I have doubled check that every error type is included. I will remove the `EasyPost.Exception` namespace from the nested list since they are not necessarily needed.

UPGRADE_GUIDE.md
jchen293: Yes, good catch

UPGRADE_GUIDE.md
jchen293: Resolved and splited the sentence to three bullet points

nwithan8 (dismissed):

UPGRADE_GUIDE.md
```diff
@@ -2,22 +2,165 @@
...
+
+*Likelihood of Impact: **High***
+
+Static and instance-based methods have been divided into separate services and resources, methods that have API interactions are in services now.
+
+For example:
+
+- The static method `Shipment.buy()` is now accessible in the Shipment service via `client.shipment.buy("shp_...")`. 
+- The instance method `myShipment.lowestRate()` is still accessible only via a Shipment instance because there is no API interaction.
+- The instance method of `refresh()` has been removed becasue you can no longer use object to make a HTTP request.
```
Justintime50, line 121:
```suggestion
- The instance method of `refresh()` has been removed because you can no longer use an object to make a HTTP request.
```

UPGRADE_GUIDE.md
```diff
@@ -2,22 +2,165 @@
...
+
+Instances of an object cannot be initialized directly. Instead, use the service to create the instance via a `create`, `retrieve`, or `all` API call.
+
+## 6.0 Removal of Setters
+
+*Likelihood of Impact: **Medium***
+
+Each object no longer has setters. The use case of the setter is extremely low, but if you need to use the setter for a specific reason, try to use `reflection` in Java.
```
Justintime50, line 129:
```suggestion
Objects no longer have setters. The use case of the setter is extremely low, but if you need to use the setter for a specific reason, try to use `reflection` in Java.
```

UPGRADE_GUIDE.md
```diff
@@ -2,22 +2,165 @@
...
+
+- Pickup class: `getPickoutRates()` -> `getPickupRates()`
+- PaymentMethod class: `getPrimaryPaymentMethodObject()` -> `getPrimaryPaymentMethod()`
+- PaymentMethod class: `getSecondaryPaymentMethodObject()` -> `getSecondaryPaymentMethod()`
+
+## 6.0 Type Changes
+
+*Likelihood of Impact: **Low***
+
+Change the below type of objects:
```
Justintime50, line 145:
```suggestion
The following properties have changed types:
```

Justintime50 (commented):
Various small nits to fix grammar/spelling

UPGRADE_GUIDE.md
jchen293: Ok, let's leave it as `Client Instance`

Justintime50 (dismissed):

Justintime50 (approved):

nwithan8 (approved):

PR #210, opened 2022-11-30
Use different endpoint for custom workflow carrier account creation (FedEx, UPS)

nwithan8:
# Description

- Carrier account creation now determines endpoint based on type

# Testing

- New unit test, cassette

# Pull Request Type

- [ ] Improvement (fixing a typo, updating readme, renaming a variable name, etc.)

### DIFF
--- a/src/main/java/com/easypost/service/CarrierAccountService.java
+++ b/src/main/java/com/easypost/service/CarrierAccountService.java
@@ -1,6 +1,7 @@
 package com.easypost.service;
 
 import com.easypost.exception.EasyPostException;
+import com.easypost.exception.General.MissingParameterError;
 import com.easypost.http.Requestor;
 import com.easypost.http.Requestor.RequestMethod;
 import com.easypost.model.CarrierAccount;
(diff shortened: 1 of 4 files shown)


jchen293 (dismissed):

Justintime50 (changes requested):
We just pushed out the release candidate, why are we making big changes like moving all the constants around?

nwithan8:
> We just pushed out the release candidate, why are we making big changes like moving all the constants around?

Mirroring the changes made when this feature was added to the C# library, we need to unify our constants under one file. Moved the exception-specific constants to the global constants file.

Justintime50:
@nwithan8 please move the exception changes to a new PR so we can keep them small and focused. The point of this PR is to address carrier account registration only.

nwithan8:
> @nwithan8 please move the exception changes to a new PR so we can keep them small and focused. The point of this PR is to address carrier account registration only.

https://github.com/EasyPost/easypost-java/pull/211

Justintime50 (approved):

src/main/java/com/easypost/service/CarrierAccountService.java
```diff
@@ -30,13 +31,18 @@ public class CarrierAccountService {
      * @return created CarrierAccount object.
      * @throws EasyPostException when the request fails.
      */
-    public CarrierAccount create(final Map<String, Object> params)
-            throws EasyPostException {
+    public CarrierAccount create(final Map<String, Object> params) throws EasyPostException {
+        String type = (String) params.get("type");
+        if (type == null) {
+            throw new EasyPostException("Carrier account type is required");
```
jchen293, line 37: You should use `MissingParameterError` here to be more explicit on the error type. There is a constant you can use `MISSING_REQUIRED_PARAMETER`

nwithan8: Resolved

jchen293 (approved):

Justintime50 (approved):

PR #213, opened 2022-12-01
Clean up exceptions

nwithan8:
# Description

We have specific exception types, but we're not using many of them (i.e. a lot of our methods throw generic `Exception` or `EasyPostException` exceptions. This PR tries to be a bit more specific with our exception throwing and docstrings, specifically in the Requestor.

- New exceptions for JSON error, encoding errors, generic HTTP errors
- Replace generic EasyPostException exceptions with specific exception types on throw statements and method signatures (more explicit)

This is UNLIKELY to be considered a breaking change, as all exceptions can still be captured by the generic `Exception` or `EasyPostException` classes. If anything, this offers more specificity for those users wanting more explicit documentation and error handling capabilities.

# Testing

- Some unit tests updated to account for change, no new tests or re-records.

# Pull Request Type


src/main/java/com/easypost/http/Requestor.java
```diff
@@ -419,11 +412,29 @@ private static EasyPostResponse makeURLConnectionRequest(final RequestMethod met
...
+     * @throws ServiceUnavailableError when the request fails due to a service unavailability.
+     * @throws GatewayTimeoutError     when the request fails due to a gateway timeout.
+     * @throws UnknownApiError         when the request fails due to an unknown API error.
      */
     public static <T> T request(final RequestMethod method, String url, final Map<String, Object> params,
-                                final Class<T> clazz, final EasyPostClient client) throws EasyPostException {
+                                final Class<T> clazz, final EasyPostClient client)
+            throws EncodingError, JsonError, RedirectError, UnauthorizedError, ForbiddenError, PaymentError,
+            NotFoundError, MethodNotAllowedError, TimeoutError, InvalidRequestError, RateLimitError,
+            InternalServerError, ServiceUnavailableError, GatewayTimeoutError, UnknownApiError, HttpError {
```
jchen293, line 437: Why was this change? `EasyPostException` should be sufficed

src/main/java/com/easypost/exception/API/ServiceUnavailableError.java
```diff
@@ -5,7 +5,7 @@
 import com.easypost.model.Error;
 import com.easypost.exception.EasyPostException;
 
-public class ServiceUnavailablError extends EasyPostException {
+public class ServiceUnavailableError extends EasyPostException {
```
jchen293, line 8: Thanks for fixing the typo, this will need to be documented in Changelog

src/main/java/com/easypost/http/Requestor.java
nwithan8: An attempt at being more explicit about what exceptions get thrown by this method, and explanations about why in the docstrings. These are all sub-types of EasyPostException, so we certainly could keep it to just EasyPostException. Like I said, just trying to be explicit so the end user knows what can be thrown, in case they want to handle different exceptions differently (e.g. simply try again on a RateLimitError, but know something is wrong for a InvalidRequestError).

Thoughts?

src/main/java/com/easypost/exception/API/ServiceUnavailableError.java
nwithan8: Resolved

src/main/java/com/easypost/http/Requestor.java
jchen293: If we already include all the possible exceptions in the Javadoc, then do we really need to include them again in the method? It just seems too much to me, won't block this.

jchen293 (approved):

Justintime50 (approved):

PR #216, opened 2022-12-07
Fix EasyVCR-related issues for unit tests

nwithan8:
# Description

- Fix unit tests disabled due to EasyVCR issues:

Summary:
- Referral credit card - The API calls to Stripe use a temporary client that was not configured to use EasyVCR. API calls to Stripe were not being recorded/replayed, and instead live data was being used. The live API calls returned data that did not match expected data in later EasyPost API calls, causing it to fail.
  - The client used for Stripe calls has been configured to use EasyVCR on testing. Additional tweaks to the URL encoding for Stripe API calls fixed HTTP issues.
- Create shipment by IDs - Creating a shipment involves creating two addresses. Both API calls included the same exact request parameters. EasyVCR is configured to record a request if it does not already exist in the cassette (only considering request parameters, not response data). The first address API call was being lost during recording, causing a data mismatch on replay.
  - This unit test has been updated to have a different parameter for the second address creation API call, ensuring that both address creation calls are recorded due to different parameters, correcting the replay issues.

# Testing

- Re-enable disabled unit tests (no more disabled unit tests)

# Pull Request Type

### DIFF
--- a/src/test/java/com/easypost/ShipmentTest.java
+++ b/src/test/java/com/easypost/ShipmentTest.java
@@ -13,7 +13,6 @@ import com.easypost.model.ShipmentCollection;
 import com.easypost.model.Smartrate;
 import com.easypost.model.SmartrateAccuracy;
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import java.util.ArrayList;
--- a/src/test/java/com/easypost/ShipmentTest.java
+++ b/src/test/java/com/easypost/ShipmentTest.java
@@ -344,14 +344,16 @@ public final class ShipmentTest {
      * @throws EasyPostException when the request fails.
      */
     @Test
-    @Disabled
-    // TODO: test is for some reason failing to pull a proper recording when playing
-    // back. Only test doing this
     public void testCreateWithIds() throws EasyPostException {
         vcr.setUpTest("create_with_ids");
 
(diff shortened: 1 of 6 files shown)


src/test/java/com/easypost/ShipmentTest.java
```diff
@@ -344,14 +343,16 @@ public void testCreateTaxIdentifiers() throws EasyPostException {
...
     public void testCreateWithIds() throws EasyPostException {
         vcr.setUpTest("create_with_ids");
 
-        Address fromAddress = vcr.client.address.create(Fixtures.caAddress1());
-        Address toAddress = vcr.client.address.create(Fixtures.caAddress1());
+        Map<String, Object> address1Data = Fixtures.caAddress1();
+        Map<String, Object> address2Data = Fixtures.caAddress2();
+        // VCR will overwrite the first address recording if the parameters are the exact same,
+        // which will cause us to lose the response from the first address creation and cause the replay to fail.
+        address2Data.put("name", "EasyPost");
```
Justintime50, line 353: Question: Why the name bit here? This should be as simple as using the 2nd address because they are totally different fixtures. We shouldn't need to mess with the data further?

nwithan8: You're right. I forgot we had two different address fixtures at our disposal, but CoPilot didn't.

Justintime50 (approved):

jchen293 (approved):

PR #220, opened 2022-12-20
add error prone and fix some potential bugs

jchen293:
# Description

Add [Error Prone](https://errorprone.info/docs/installation) to the library and fix the below errors.
Modified the CI build to use JDK 19 to build against java 8 - 19, in order to compatible with Error Prone

Screen Shot 2022-12-20 at 4 25 02 PM
Screen Shot 2022-12-20 at 4 35 49 PM

# Testing

Unit tests pass

# Pull Request Type


.mvn/jvm.config
```diff
@@ -0,0 +1,10 @@
+--add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
```
Justintime50, line 1: Question: What is all this?

src/main/java/com/easypost/Constants.java
```diff
@@ -18,21 +18,21 @@ public abstract class Constants {
...
         public static final String NO_OBJECT_FOUND = "No %s found.";
-        public static final String NO_PAYMENT_METHODS =
-                "No payment methods are set up. Please add a payment method and try again.";
+        public static final String NO_PAYMENT_METHODS = "No payment methods are set up. Please add a payment"
+                + " method and try again.";
         public static final String API_DID_NOT_RETURN_ERROR_DETAILS = "API did not return error details.";
-        public static final String WEBHOOK_DOES_NOT_MATCH =
-                "Webhook received did not originate from EasyPost or had a webhook secret mismatch.";
+        public static final String WEBHOOK_DOES_NOT_MATCH = "Webhook received did not originate from EasyPost or had a"
+                + " webhook secret mismatch.";
```
Justintime50, line 35: Let's keep the spaces on the beginning/end of the strings consistent instead of having both with the plus sign.

src/test/java/com/easypost/TestUtils.java
```diff
@@ -28,37 +28,20 @@ public enum ApiKey {
...
+    private static final List<String> QUERY_CENSORS = ImmutableList.of("card[cvc]", "card[number]");
+    private static final List<String> BODY_CENSORS = ImmutableList.of(
+            "api_keys",
+            "client_ip",
+            "credentials",
+            "key",
+            "keys",
+            "phone_number",
+            "phone",
+            "test_credentials");
```
Justintime50, line 41: Can we move the ending parenthesy here to the next line?

.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        javaversion: [ "8", "9", "10", "11", "12", "13", "14", "15", "16", "17", "18", "19" ]
+        javaversion: [ "11", "12", "13", "14", "15", "16", "17", "18", "19" ]
```
Justintime50, line 14: Whoa... we need to build against Java 8 heh. We support it. Do we need to use an older version of Error Prone that supports it?

jchen293: That's what I am trying to figure out, Java VM was not able to create on Java 8 and it seems like it needs at least version 11 to create it: https://stackoverflow.com/questions/72045938/sonarqube-could-not-create-the-java-virtual-machine-unrecognized-option-add-e

.mvn/jvm.config
jchen293: This is part of the installation guide for Maven: https://errorprone.info/docs/installation

jchen293:
Closing this PR due to incompatibility with Java 8 for `--add-exports` which is required for the Error Prone installation setup for Maven. I will reopen this PR when we drop support for Java 8 in the future.

Justintime50:
Let's take another look at this after the holidays. Often there is something to be done.

nwithan8:
Vulnerabilities being suppressed are due to an outdated dependency that ErrorProne uses.
ErrorProne is a plugin we use during compilation; it is not a dependency of our source code.
The vulnerability does not extend to our compiled JAR.

nwithan8 (dismissed):

.github/workflows/ci.yml
```diff
@@ -23,12 +23,32 @@ jobs:
             ${{ runner.os }}-maven-
       - name: Install dependencies
         run: make install
+      - name: Set up JDK 19 for compilation
```
Justintime50, line 26: Would prefer to keep these description lines static, we'll forget to update these transient numbers

```suggestion
      - name: Set up JDK for compilation
```

.github/workflows/ci.yml
```diff
@@ -23,12 +23,32 @@ jobs:
             ${{ runner.os }}-maven-
```
Justintime50, line 23: We discussed cutting out all the non-LTS versions of Java in the matrix above, we still doing that?

.github/workflows/ci.yml
```diff
@@ -23,12 +23,32 @@ jobs:
...
-      - name: Build and test with Maven
+      - name: Run test with Java ${{ matrix.javaversion }}
+        run: EASYPOST_TEST_API_KEY=123 EASYPOST_PROD_API_KEY=123 make test
+  coverage:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install dependencies
+        run: make install
+      - name: Set up JDK 19 for compilation
```
Justintime50, line 46:
```suggestion
      - name: Set up JDK for compilation
```

dependency-check-suppressions.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
+    <suppress>
+        <vulnerabilityName>CVE-2022-3171</vulnerabilityName>
```
Justintime50, line 4: Can we include a comment here why we are ignoring these? (Error Prone issues)

pom.xml
```diff
@@ -196,10 +203,39 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <version>3.9.0</version>
+                <version>3.8.0</version>
```
Justintime50, line 206: Question: why did we downgrade?

run-error-prone.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
Justintime50, line 1: This script shouldn't be needed anymore since we got the plugin to work instead. Let's delete this.

src/main/java/com/easypost/Constants.java
Justintime50: This still needs to be addressed ^

src/test/java/com/easypost/TestUtils.java
Justintime50: This still needs to be addressed

Justintime50 (changes requested):

.github/workflows/ci.yml
jchen293: We are using JDK 19 to test Java 8 - 19 version, we still need to keep the matrix for the build

pom.xml
jchen293: Reverted, I was following the Error Prone guide.

Justintime50 (approved):

nwithan8 (approved):

PR #224, opened 2023-01-11
Drop vulnerable Apache dependencies, implement dependent functions

nwithan8:
# Description

- Extract required Apache functions (hex conversion) into ApachePatch
- Drop vulnerable Apache dependencies

There were two Apache dependencies being imported into this library, `httpcomponents` (explicitly) and `common-codec` (implicitly) both of which were vulnerable.

Turns out `httpcomponents` was not being used anywhere, and could simply be dropped.

`common-codec` was being used solely for its Hex class, used for hexidecimal conversion for webhook validation. The necessary functions from that class have been extracted into a `ApachePatch` internal utility file, allowing us to drop the vulnerable dependency forever. This, of course, will not stop `common-codec` from being imported into our library via the ErrorProne plugin, but will at least ensure we are not utilizing the vulnerable package.

# Testing

All unit tests pass, particularly webhook validation (the only test that would be impacted by the change), suggesting that the drop-in replacement will not introduce a breaking change.

# Pull Request Type

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,11 +77,6 @@
             <scope>compile</scope>
         </dependency>
         <dependency>
-            <groupId>org.apache.httpcomponents</groupId>
-            <artifactId>httpclient</artifactId>
-            <version>4.5.13</version>
-        </dependency>
-        <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
             <version>1.18.24</version>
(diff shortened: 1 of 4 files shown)


jchen293:
@nwithan8 Can you provide a screenshot of the vulnerability? On the Maven website, the version we are currently using `4.5.13` does not have any vulnerabilities: https://mvnrepository.com/artifact/org.apache.httpcomponents/httpclient

nwithan8:
> @nwithan8 Can you provide a screenshot of the vulnerability? On the Maven website, the version we are currently using `4.5.13` does not have any vulnerabilities: https://mvnrepository.com/artifact/org.apache.httpcomponents/httpclient

This package wasn't vulnerable, but wasn't actually used by anything, so it can be dropped.

jchen293 (approved):

PR #230, opened 2023-01-21
Drop automagical endpoint calculation, replace with explicit endpoints

nwithan8:
# Description

I've never been a fan of how this (and some of our other libraries) determine what endpoint to hit in our API by basically taking the name of the class, figuring out how to make it plural and then lowercasing it. The "class URL" and "instance URL" utility functions are clunky, do unnecessary reflection, are hard to follow, and are simply begging to break one day in the future simply because a class name does not one-to-one match an endpoint on the API.

Instead, as was done in the C# library during its rewrite last year, this "automagical" utility was removed and instead replaced with explicit endpoint strings. This not only returns transparency to what endpoint is being hit on each function, but also dramatically improves readability of our code.

This PR also makes minor tweaks to how the requestor function operates under-the-hood, mainly by containing client API base and version information solely to that class. Previously, the whole URL (including base and API version) were determined in each service. Now, that calculation is pushed downstream into one location inside the actual request function. This restructuring also allows for the introduction of API version override on a per-request basis, making "beta" endpoints more obvious and meaning less changes needed when a beta features migrates to general availability and no longer needs to hit the "beta" API version.

- Remove endpoint determination by class, replace with explicit endpoint strings
- Modify Requestor under-the-hood to handle new endpoint strings
- Add support for overriding API version on a per-request basis using new endpoint strings

# Testing

- All unit tests pass with no re-recording necessary, implying that the change does not introduce a breaking change to the end-user.
  - Modifications were needed to the mocking rules for Billing unit tests to account for the changes made to the request function downstream. This did not affect the actual functionality or assertions of those unit tests.

# Pull Request Type


src/main/java/com/easypost/http/Requestor.java
```diff
@@ -430,14 +429,51 @@ private static EasyPostResponse makeURLConnectionRequest(final RequestMethod met
...
+     * @throws MethodNotAllowedError   when the request method is not allowed.
+     * @throws TimeoutError            when the request times out.
+     * @throws InvalidRequestError     when the request is invalid.
+     * @throws RateLimitError          when the request exceeds the rate limit.
+     * @throws InternalServerError     when the request fails due to an internal server error.
+     * @throws ServiceUnavailableError when the request fails due to a service unavailability.
+     * @throws GatewayTimeoutError     when the request fails due to a gateway timeout.
+     * @throws UnknownApiError         when the request fails due to an unknown API error.
+     */
+    public static <T> T request(final RequestMethod method, final String endpoint, final Map<String, Object> params,
```
jchen293, line 469: I am fine with this PR. The only concern I have is this public function because it takes in the `apiVersion` parameter, meaning that users can use this function with any `apiVersion`. I believe people will probably not use this `request` function anyway but just wanted to bring this up.

src/main/java/com/easypost/service/BatchService.java
```diff
@@ -97,9 +92,7 @@ public Batch createAndBuy(final Map<String, Object> params) throws EasyPostExcep
      * @throws EasyPostException when the request fails.
      */
     public Batch label(final String id, final Map<String, Object> params) throws EasyPostException {
-        String url = String.format("%s/label", InternalUtilities.instanceURL(Batch.class, id));
-
-        return Requestor.request(RequestMethod.POST, url, params, Batch.class, client);
+        return Requestor.request(RequestMethod.POST, "batches/" + id + "/label", params, Batch.class, client);
```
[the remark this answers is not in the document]
nwithan8: Cleaner vs. waste of a variable, I'll defer to you.

src/main/java/com/easypost/http/Requestor.java
nwithan8: Fair enough, I'll take a crack at making this function inaccessible to end-users since, you're right, no one should be manually calling this.

nwithan8: I don't think this will be a problem, since, as you said, users shouldn't ever have a need to call this method. But if they did, they could technically hit a different endpoint, as well as a different API version. What I'm saying is, the additional API version parameter doesn't introduce any more risk or anti-pattern than was already present.

I would love for this to be all internal-only access, but as we've discussed before, this method has to be public since it is in a different package/namespace than the services. Such is Java.

src/main/java/com/easypost/service/BatchService.java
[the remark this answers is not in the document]
Justintime50: I won't block on it, but I don't think it's actually cleaner since you're cramming string interpolation into a function call instead of only giving it finalized parameters.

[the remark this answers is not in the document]
nwithan8: Resolved

src/main/java/com/easypost/http/Requestor.java
jchen293: We should avoid this if we could. One way to solve this is by passing another client with `Beta` apiVersion when initializing `BetaReferralCustomerService` in the EasyPostClient class constructor. This way, we can remove this function.

nwithan8: Yeah, we should do it that way, I believe that was part of the spec for our rewrite design on how to handle beta functionality, but I'm not seeing a `BetaClient` in the library (a la C#: https://github.com/EasyPost/easypost-csharp/blob/master/EasyPost/BetaClient.cs)

Thoughts @Justintime50 ?

Justintime50: My suggestion would be to take the path of least resistance. In PHP for instance, we simply have a beta flag on the requestor function that tells it to hit the beta endpoint instead. I would recommend going a similar route.

nwithan8: Good, so what we have here is good to go.

Justintime50 (approved):
This is a classic "If it ain't broke, don't fix it" example. I'm not convinced we got value out of the these changes - they also don't really hurt to have either. Seems fine to me.

I would caution though that what efforts we choose to pursue and code that gets refactored should be focused on what returns the most value to us and especially end-users. This is fine, though now fragments how our libs handle endpoint params and wasn't explicitly necessary since the previous implementation has worked great for the last ~8 years.

jchen293 (approved):

PR #232, opened 2023-01-30
add retrieve stateless rates function

jchen293:
# Description

add retrieve stateless rates function
add retrieve the lowest rate of stateless rates function

# Testing

add two unit tests

# Pull Request Type

### DIFF
--- a/src/main/java/com/easypost/model/StatelessRateDeserializer.java
+++ b/src/main/java/com/easypost/model/StatelessRateDeserializer.java
@@ -1,0 +1,31 @@
+package com.easypost.model;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+
+import java.lang.reflect.Type;
+
+public final class StatelessRateDeserializer implements JsonDeserializer<StatelessRate[]> {
+    /**
+     * Deserialize a StatelessRate from a JSON object.
+     *
+     * @param json    JSON object to deserialize.
+     * @param typeOfT Type of the object to deserialize.
+     * @param context Deserialization context.
+     * @return Deserialized StatelessRate object.
+     * @throws JsonParseException if the JSON object is not a valid StatelessRate.
+     */
+    @Override
+    public StatelessRate[] deserialize(final JsonElement json, final Type typeOfT,
+    final JsonDeserializationContext context) throws JsonParseException{
+        JsonObject jo = json.getAsJsonObject();
+        JsonElement results = jo.get("rates");
+        Gson gson = new Gson();
+
+        return gson.fromJson(results, StatelessRate[].class);
+    }
+}
--- a/src/main/java/com/easypost/utils/Utilities.java
+++ b/src/main/java/com/easypost/utils/Utilities.java
@@ -51,7 +51,7 @@ public abstract class Utilities {
 
         if (lowestRate == null) {
             throw new FilteringError(String.format(
-                    Constants.ErrorMessages.NO_OBJECT_FOUND, "lowest rate matching required criteria"));
+                    Constants.ErrorMessages.NO_OBJECT_FOUND, "rates"));
         }
 
         return lowestRate;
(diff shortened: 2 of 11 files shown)


src/main/java/com/easypost/model/StatelessRateDeserializer.java
```diff
@@ -0,0 +1,31 @@
...
+import com.google.gson.Gson;
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+
+import java.lang.reflect.Type;
+
+public final class StatelessRateDeserializer implements JsonDeserializer<StatelessRate[]> {
```
nwithan8, line 12: Question: Why do we need a custom deserializer for Stateless Rate? Couldn't we just deserialize this to a collection, and then have the function itself return the `rates` property of that collection, rather than effectively overriding the deserialization process to start deserialization at the "rates" key?

I would imagine this is similar to the SmartRate deserialization, which also has a custom deserializer. I guess refresh my memory as to why we didn't get rid of that a few months ago? (I know I was the one who added it, but that was the first week I was here, I didn't know what I know now).

nwithan8 (changes requested):

src/main/java/com/easypost/model/StatelessRateDeserializer.java
jchen293: Because the API not only returns a list of rates but also `from_address`, `to_address`, `parcel`, and `message`, so we need some tweaks for the deserialization.

Yes, creating a collection works but we will need to create another class for `option` because there isn't one. As a result, I went for a custom deserializer because we don't need to introduce two additional classes in the model and we are only deserializing what we needed.

nwithan8: Fair enough, I just don't want to get to a point where we have a half-dozen custom deserializers. This is acceptable.

nwithan8 (dismissed):

src/main/java/com/easypost/utils/Utilities.java
```diff
@@ -8,13 +8,55 @@
...
+            }
+
+            if (lowestRate == null || Float.parseFloat(rate.getRate()) < Float.parseFloat(lowestRate.getRate())) {
+                lowestRate = rate;
+            }
+        }
+
+        if (lowestRate == null) {
+            throw new FilteringError(String.format(
+                    Constants.ErrorMessages.NO_OBJECT_FOUND, "lowest rate matching required criteria"));
```
Justintime50, line 54: This error message doesn't follow the previously defined pattern of `No rates found`. Let's make sure this matches the rest of the lowest rate error messages

src/test/java/com/easypost/BetaStatelessRateTest.java
```diff
@@ -0,0 +1,61 @@
+package com.easypost;
```
Justintime50, line 1: This file needs to be renamed to `BetaRateTest` since all test files must match the lib file and eventually this could contain more than just a stateless rate.

src/test/java/com/easypost/BetaStatelessRateTest.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.List;
+
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import com.easypost.exception.EasyPostException;
+import com.easypost.model.StatelessRate;
+import com.easypost.utils.Utilities;
+
+public class BetaStatelessRateTest {
```
Justintime50, line 16: Please adjust the class name with the file name

src/test/java/com/easypost/BetaStatelessRateTest.java
```diff
@@ -0,0 +1,61 @@
...
+    @Test
+    public void testRetrieveLowestStatelessRate() throws EasyPostException {
+        vcr.setUpTest("retrieve_lowest_stateless_rate");
+
+        HashMap<String, Object> shipment = Fixtures.basicShipment();
+
+        List<StatelessRate> rates = vcr.client.betaRate.retrieveStatelessRates(shipment);
+        StatelessRate lowestRate = Utilities.getLowestStatelessRate(rates, null, null);
+
+        assertEquals("First", lowestRate.getService());
```
Justintime50, line 59: If it's simple to do, please also test the error messages for this function

Justintime50 (changes requested):

jchen293:
Forced push to rebase

src/test/java/com/easypost/BetaStatelessRateTest.java
jchen293: Added

Justintime50 (approved):

nwithan8 (approved):

PR #237, opened 2023-02-17
Remove extraneous Apache port functions to bump coverage percentage

nwithan8:
# Description

Removes a bunch of overloaded methods ported from the Apache hex library, only keep what we need.

This will reduce the number of uncovered lines significantly.

# Testing

# Pull Request Type


Justintime50 (approved):
All looks like stuff we don't/won't use, ship it.

Please do this in EasyVCR too

Justintime50:
We should have bumped the minimum coverage number with this too

Issue #239, opened 2023-03-15
[Bug]: IllegalStateException when deseralizing "error"


terrettazmaxime:
### Software Version

6.3.0

### Language Version

11

### Operating System

Linux

### What happened?

We have noticed a bug in the software that occurs every time there is a validation errors when performing a One-Call Buy Shipment with Purolator and they return multiple validation errors. 
The bug is related to the deserialization of the JSON errors, which is not functioning as expected.
Specifically, we have identified that when your code is deserializing the error node, it expects the message to be either a string or an array, but in this case, it is an object, it cashes in your internal code, and we do not get an EasyPostException.
This issue is causing us significant trouble as we are currently unable to move forward with our production process.

I have included the JSON that is causing the problem below, as an example:
```
{
    "error": {
        "code": "UNPROCESSABLE_ENTITY",
        "message": {
            "errors": [
                "1100156: From On Label address is an invalid Service Point.",
                "1100674: Postal Code XXX is only valid for XXX."
            ]
        },
        "errors": [
            {
                "message": "1100156: From On Label address is an invalid Service Point."
            },
            {
                "message": "1100674: Postal Code XXX is only valid for XXX."
            }
        ]
    }
}
```

Code of your Error class
package com.easypost.model;

```
import java.util.List;

public final class Error {
    private String message;
    private String code;
    private List<Error> errors;
    private String suggestion;
    private String field;

    public Error() {
    }

    void setMessage(String message) {
        this.message = message;
    }

    void setCode(String code) {
        this.code = code;
    }

    public String getMessage() {
        return this.message;
    }

    public String getCode() {
        return this.code;
    }

    public List<Error> getErrors() {
        return this.errors;
    }

    public String getSuggestion() {
        return this.suggestion;
    }

    public String getField() {
        return this.field;
    }
}
```

When your deserializer, do the work, the message node is not a string, neither an array but and object and it crash.
```
public final class ErrorDeserializer implements JsonDeserializer<Error> {
    public ErrorDeserializer() {
    }

    public Error deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) throws JsonParseException {
        JsonObject jo = json.getAsJsonObject();
        JsonElement results = jo.get("error");
        Gson gson = new Gson();
        if (results == null) {
            Error error = new Error();
            error.setMessage("API did not return error details.");
            error.setCode("NO RESPONSE CODE");
            return error;
        } else {
[...]

PR #240, opened 2023-03-15
Add support for jsonElement in error deserialization

terrettazmaxime:
# Description

Adding correction for issue #239 

# Testing

# Pull Request Type


jchen293 (changes requested):
Thank you for making this pull request!

While this change would address the current edge case, we want to enhance the logic to cover various error scenarios that may not follow the expected field types. This will help us safeguard against potential edge cases that may arise in the future.

Can we adjust this to first try to stringify the object and if that fails, return a default error message stating the error could not be deserialized? This message could also suggest looking in the `errors` field as that will often contain additional details when the `message` field cannot be deserialized properly.

jchen293:
Per our conversation: https://github.com/EasyPost/easypost-java/issues/239#issuecomment-1476846855
I am going to close this PR, please see #242 PR for reference

PR #242, opened 2023-03-20
Handle edge cases when deserializing error JSON

### LINKED ISSUE #239: [Bug]: IllegalStateException when deseralizing "error"

jchen293:
# Description

Handle edge cases when deserializing error JSON, close #239 

# Testing

Add two unit tests and pass

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v6.4.0 Next Release
+
+- Handle edge cases when deseralizing error JSON, close Github issue [239](https://github.com/EasyPost/easypost-java/issues/239).
+
 ## v6.3.0 (2023-02-21)
 
 - Adds beta `retrieveStatelessRates` function to get stateless rates
--- a/src/main/java/com/easypost/model/ErrorDeserializer.java
+++ b/src/main/java/com/easypost/model/ErrorDeserializer.java
@@ -12,8 +12,32 @@ import com.google.gson.JsonPrimitive;
 
 import java.lang.reflect.Type;
 import java.util.ArrayList;
+import java.util.Map.Entry;
 
 public final class ErrorDeserializer implements JsonDeserializer<Error> {
+    /**
+     * Recursively traverse an error JSON element and its sub-element(s), and extracts all
+     * error string values found into the specified string list.
+     *
+     * @param element the JSON element to traverse
+     * @param messagesList the list of strings to append found values to
+     */
+    private void traverseJsonElement(JsonElement element, ArrayList<String> messagesList) {
+        if (element.isJsonPrimitive()) {
+            messagesList.add(element.getAsString());
+        } else if (element.isJsonArray()) {
+            JsonArray array = element.getAsJsonArray();
+            for (JsonElement arrayElement : array) {
+                traverseJsonElement(arrayElement, messagesList);
+            }
+        } else if (element.isJsonObject()) {
+            JsonObject object = element.getAsJsonObject();
+            for (Entry<String, JsonElement> entry : object.entrySet()) {
+                traverseJsonElement(entry.getValue(), messagesList);
+            }
+        }
+    }
+
     /**
      * Deserialize an Error from a JSON object.
      *
@@ -25,7 +49,7 @@ public final class ErrorDeserializer implements JsonDeserializer<Error> {
      */
     @Override
     public Error deserialize(final JsonElement json, final Type typeOfT,
-    final JsonDeserializationContext context) throws JsonParseException{
+    final JsonDeserializationContext context) throws JsonParseException {
         JsonObject jo = json.getAsJsonObject();
         JsonElement results = jo.get("error");
         Gson gson = new Gson();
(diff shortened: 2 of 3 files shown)


src/main/java/com/easypost/model/ErrorDeserializer.java
```diff
@@ -12,8 +12,32 @@
...
 
 public final class ErrorDeserializer implements JsonDeserializer<Error> {
+    /**
+     * Recursively traverse an error JSON element and its sub-element(s), and extracts all
+     * error string values found into the specified string list.
+     *
+     * @param element the JSON element to traverse
+     * @param messagesList the list of strings to append found values to
+     */
+    private void traverseJsonElement(JsonElement element, ArrayList<String> messagesList) {
```
nwithan8, line 25: This works because Java is "pass by reference", so the `messageList` variable getting updated by each recursive call does reflect on the variable on line 65 below.

However, this won't work in other languages that don't pass by reference. Instead, can we make this recursive function actually return the updated/appended `messageList` variable so we don't have to worry about its reference?

nwithan8 (changes requested):
See comment above.

src/main/java/com/easypost/model/ErrorDeserializer.java
jchen293: This will work with the other six languages as they support `pass by reference` too or has `pass by reference` already for passing in an object into a method, some languages might require you to add an additional symbol, like in Go you would need to use a pointer, and in PHP you would need to use `&`.

nwithan8 (changes requested):

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v6.4.0 Next Release
```
Justintime50, line 3: Nit: version numbers are picked out at release time based on what's included, this may not be the only thing changed, for example this is really a patch and not a feature so it will most likely be 6.3.1. Let's not pre-label versions in the future

src/main/java/com/easypost/model/ErrorDeserializer.java
```diff
@@ -12,8 +12,32 @@
...
 import java.util.ArrayList;
+import java.util.Map.Entry;
 
 public final class ErrorDeserializer implements JsonDeserializer<Error> {
+    /**
+     * Recursively traverse an error JSON element and its sub-element(s), and extracts all
+     * error string values found into the specified string list.
+     *
+     * @param element the JSON element to traverse
+     * @param messagesList the list of strings to append found values to
```
Justintime50, line 23: question: this description doesn't make sense, the last 3 words seem to not belong here or are unfinished?

src/main/java/com/easypost/model/ErrorDeserializer.java
```diff
@@ -36,18 +60,18 @@ public Error deserialize(final JsonElement json, final Type typeOfT,
...
-            JsonArray jsonArray = errorMessage.getAsJsonArray();
-            ArrayList<String> messages = new ArrayList<>();
-
-            for (int i = 0; i < jsonArray.size(); i++) {
-                messages.add(jsonArray.get(i).getAsString());
-            }
 
+        try {
+            ArrayList<String> messages = new ArrayList<>();
+            JsonElement errorMessageJson = results.getAsJsonObject().get("message");
```
Justintime50, line 66: question: how does this work if it's a list? We are trying to get it as an object so will that mean we no longer capture when something is a list?

nwithan8: It will take the value of the "message" key and turn it into a `JsonObject`, the root of any JSON element for the GSON library. In the `traverseJsonElement` function, it will handle if it's a list/array.

src/main/java/com/easypost/model/ErrorDeserializer.java
jchen293: `specified string list` here is the `messagesList`

src/main/java/com/easypost/model/ErrorDeserializer.java
jchen293: Nate is correct, everything is being handled in the `traverseJsonElement` function recursively.

src/main/java/com/easypost/model/ErrorDeserializer.java
Justintime50: mmm re-reading this I just misunderstood it

nwithan8 (approved):

PR #269, opened 2023-05-18
[chore] Explicitly define "Accept" header for requests

nwithan8:
# Description

Addresses bug in Java 8 which uses a confusing default "Accept" header if not defined: https://bugs.openjdk.org/browse/JDK-8163921

Not a breaking change, but makes our API logs easier to read and consistent with expected behavior.

# Testing

- N/A. Does not affect unit tests or cassettes since we do not enforce matching by headers in our VCR.

# Pull Request Type


jchen293 (approved):

src/main/java/com/easypost/http/Requestor.java
```diff
@@ -371,6 +371,7 @@ private static EasyPostResponse makeURLConnectionRequest(final RequestMethod met
                             String.format("Unrecognized HTTP method %s. Please contact EasyPost at %s.", method,
                                     Constants.EASYPOST_SUPPORT_EMAIL));
             }
+            conn.setRequestProperty("Accept", "*/*");
```
Justintime50, line 374: issue: We should only be accepting JSON back from the API because we only know how to handle JSON in the lib. The other libs are set to JSON as well

Justintime50 (changes requested):
Per convos on Slack, let's hold off on this for a bit

Justintime50:
This probably isn't needed anymore

PR #275, opened 2023-06-15
[chore] Mock certain API responses in unit tests

nwithan8:
# Description

Some of our unit tests can benefit from mocking:
- Billing unit tests are currently mocked using Mockito, which overrides an entire set of nested function calls. This PR substitutes this with a new mocking utility that only mocks the HTTP request aspect of the tested function, allowing true execution of other aspects (request prep and response handling) of the function stack.
- Rate filtering (lowest rate, lowest SmartRate) utilizes real data from the API. Depending on the day when these cassettes are re-recorded, the lowest rate returned by the API call may differ, requiring devs to update the unit test assertions. By mocking the API call, we can still test the local filtering function (what is actually being tested), but with consistent data without needing to adjust assertions.

This PR:
- Introduces a new `MockClient` class that, rather than executing a real HTTP request, finds a matching configured request (via HTTP method and endpoint regex) and returns a configured response (status code and content)
- Reworks stacktrace (client -> service -> requestor is now client -> service -> client -> requestor), allowing MockClient to inject itself into the flow via override to execute as required.
  - A publicly-accessible `request` function now exists on an `EasyPostClient` instance. End-users could use this to manually call any endpoint with raw parameters rather than calling it via a specific function (opens up extensibility)
  - **No breaking changes introduced for end-users**
- Add mocking capabilities to the test suite's VCR utility. When provided with mocked requests, the VCR client for any given unit test will first attempt to return one of the configured mock requests, falling back to the typical VCR if not found. This allows users to continue to use VCR as usual, but mock only certain requests (e.g. mock one specific API request, but use the normal VCR for the other five in this particular unit test).
- Implement mocking for billing and rate filtering-related unit tests

# Testing

- Manually confirmed that mock client is being executed via override as expected
- Mock requests are not recorded to cassette (since they do not make it to the VCR stage). Technically some cassettes now have unneeded recording entries, but cassettes have not been re-recorded.

# Pull Request Type


src/main/java/com/easypost/service/EasyPostClient.java
```diff
@@ -189,4 +208,41 @@ public String getApiVersion() {
...
+     * Execute an HTTP request for a given method and URL with this EasyPostClient object.
+     *
+     * @param method   the HTTP method to use
+     * @param endpoint the API endpoint to use
+     * @param params   the parameters to use
+     * @param clazz    the class to use for deserializing the response
+     * @param <T>      the type of object to return
+     * @return the deserialized response
+     */
+    public <T> T request(final Requestor.RequestMethod method, final String endpoint, final Map<String, Object> params,
```
jchen293, line 222: Question: why are we moving the `request` functions from `Requestor.java` to `EasyPostClient.java`? Do you need to make this change to create the `MockClient` class?

src/test/java/com/easypost/BetaRateTest.java
```diff
@@ -34,7 +73,11 @@ public static void setup() throws EasyPostException {
      */
     @Test
     public void testRetrieveStatelessRates() throws EasyPostException {
-        vcr.setUpTest("retrieve_stateless_rates");
+        List<MockRequest> mockRequests = new ArrayList<>();
+        mockRequests.add(new MockRequest(new MockRequestMatchRules(Requestor.RequestMethod.POST, ".*\\/rates.*"),
+                new MockResponse(200, retrieveRatesResponseJson)));
+
+        vcr.setUpTest("retrieve_stateless_rates", mockRequests);
```
jchen293, line 80: Question: with the mocking, it still makes the real API request and records to a cassette but is using the mock data instead of the API response?

nwithan8: No, any mock request is not actually executed against the API.

jchen293: If this request is mocked which is not hitting the API endpoint, why do we need `vcr.setUpTest` and can we remove the existing corresponding cassettes (if there is one)?

nwithan8: Sure. Wanted to get some feedback on this anyway, since we could keep the VCR structure and pass in mock requests to it (allowing us to do a combination of mock and real requests), or remove the VCR and use a mock client instead (allowing only mock requests). For tests that only make one API call, it can go either way.

Thoughts?

jchen293: We probably don't need both ways. I would remove the VCR and use a mock client, this seems the easier way for any endpoints that we need to mock

Justintime50: Mocking, (as we've previously discussed at length), should be used as a last resort. It doesn't execute real code, it doesn't make real API calls, and it's very prone to becoming out of date or making things messy (ala ~30+ lines above of concatenated strings.

We should be making real API calls and using VCR unless it's out of the question (eg: billing tests)

src/test/java/com/easypost/OrderTest.java
```diff
@@ -169,7 +173,57 @@ public void testNewRate() throws EasyPostException {
      */
     @Test
     public void testLowestRate() throws EasyPostException {
-        vcr.setUpTest("lowest_rate");
+        final String orderResponseJson = "{\n" + "  \"rates\": [\n" + "    {\n" +
```
Justintime50, line 176: thought: I'm not convinced this is an improvement. Having long concatenated strings like this incredibly messy and difficult to deal with.

nwithan8: The mock client can also take in an Object and serialize it to JSON if you want something cleaner.

Justintime50: I'd take objects over strings every day of the week. I'd also suggest trimming this down to the bare minimum we need to get the test working (no need to include a bunch of extra hard-coded data when it's not needed)

src/test/java/com/easypost/BetaRateTest.java
Justintime50: Options:

1. change the assertions (we should have been asserting that the rate field is a float instead of an exact value) - this problem should also get better after July 9 once those USPS rates adjust (the service name specifically). May honestly be more concrete
2. only mock the data we absolutely need in its most simple form

nwithan8: Correct. This attempts to do Option 2 (and will be cleaned up)

src/main/java/com/easypost/service/EasyPostClient.java
```diff
@@ -189,4 +208,41 @@ public String getApiVersion() {
...
+                         final Class<T> clazz)
+            throws GatewayTimeoutError, RateLimitError, InvalidRequestError, NotFoundError, TimeoutError, EncodingError,
+            UnauthorizedError, MethodNotAllowedError, InternalServerError, UnknownApiError, ServiceUnavailableError,
+            ForbiddenError, JsonError, HttpError, RedirectError, PaymentError {
```
jchen293, line 226: We should either just throw `EasyPostException` here since it's a parent class of all the Errors here. If we decide to keep it, you need to add the docstring for all the exceptions. This also apply to below overload function too

README.md
```diff
@@ -146,6 +146,29 @@ This project uses [EasyVCR](https://github.com/EasyPost/easyvcr-java) to record
 
 **Test Data:** The test suite has been populated with various helpful fixtures that are available for use, each completely independent from a particular user **with the exception of the USPS carrier account ID** (see [Unit Test API Keys](#unit-test-api-keys) for more information) which has a fallback value of our internal testing user's ID. Some fixtures use hard-coded dates that may need to be incremented if cassettes get re-recorded (such as reports or pickups).
 
+For some unit tests, you may need to mock certain API calls. The VCR utility in the test suite accepts an option list of `MockRequest` objects via the `setUpTest` function (see [Mocking](#mocking) below).
```
Justintime50, line 149: thought: would prefer a subheading here like the others title `Mocking`

Lul now I see the mocking section below. Why not have this be under that? Seems weird to find info about one thing spread across two places

src/test/java/com/easypost/BetaRateTest.java
```diff
@@ -34,7 +51,11 @@ public static void setup() throws EasyPostException {
      */
     @Test
     public void testRetrieveStatelessRates() throws EasyPostException {
-        vcr.setUpTest("retrieve_stateless_rates");
+        List<MockRequest> mockRequests = new ArrayList<>();
+        mockRequests.add(new MockRequest(new MockRequestMatchRules(Requestor.RequestMethod.POST, ".*\\/rates.*"),
+                new MockResponse(200, MOCK_STATELESS_RATE_RESPONSE)));
+
+        vcr.setUpTest("retrieve_stateless_rates", mockRequests);
```
jchen293, line 58: We are not testing the internal function `lowest_rate` here, we should hit the API endpoint and shouldn't really mock this test

README.md
```diff
@@ -158,3 +181,31 @@ Some tests may require an EasyPost user with a particular set of enabled feature
 - `USPS_CARRIER_ACCOUNT_ID` (eg: one-call buying a shipment for non-EasyPost employees)
 - `PARTNER_USER_PROD_API_KEY` (eg: creating a referral user)
 - `REFERRAL_CUSTOMER_PROD_API_KEY` (eg: adding a credit card to a referral user)
+
+#### Mocking
+
+We have implemented a custom, lightweight HTTP mocking functionality in this library that allows us to mock HTTP calls and responses.
```
Justintime50, line 187: nit: the we/us bit in here isn't necessary. It may read better if stated "there is a lightweight"...

src/test/java/com/easypost/BetaRateTest.java
```diff
@@ -17,6 +25,15 @@
 public class BetaRateTest {
     private static TestUtils.VCR vcr;
 
+    private static final MockStatelessRateResponse MOCK_STATELESS_RATE_RESPONSE = new MockStatelessRateResponse(
+            ImmutableList.of(
+                    new MockRate("6.07", "USPS", "First"),
```
Justintime50, line 30: praise: this is better, thank you

thought: we may want to get ahead of the service level issue now since this is mocked? We could fix the `First` and `ParcelSelect` now while we are in here

src/test/java/com/easypost/BillingTest.java
```diff
@@ -1,46 +1,25 @@
...
-    @AfterAll
-    public static void cleanup() {
-        requestMock.close();
-    }
+    private final String retrievePaymentMethodResponseJson =
+            "{\"id\":\"cust_...\",\"object\":\"PaymentMethods\",\"primary_" +
+                    "payment_method\":{\"id\":\"card_...\",\"disabled_at\":null,\"object\":\"CreditCard\",\"na" +
+                    "me\":null,\"last4\":\"4242\",\"exp_month\":1,\"exp_year\":2025,\"brand\":\"Visa\"},\"secondary_" +
+                    "payment_method\":{\"id\":\"card_...\",\"disabled_at\":null,\"object\":\"CreditCard\",\"name\":nu" +
+                    "ll,\"last4\":\"4444\",\"exp_month\":1,\"exp_year\":2025,\"brand\":\"Mastercard\"}}";
```
jchen293, line 22: Can we format this string to something like [this](https://github.com/EasyPost/easypost-java/blob/master/src/test/java/com/easypost/ErrorTest.java#L176-%20L195) that I have previously done for error test?

src/test/java/com/easypost/mocking/MockClient.java
```diff
@@ -0,0 +1,217 @@
...
+     * @param clazz      The class to deserialize the response to.
+     * @param apiVersion The API version to use.
+     * @param <T>        The type of the response.
+     * @return The deserialized response.
+     */
+    private <T> T processMockRequest(Requestor.RequestMethod method, String endpoint, Map<String, Object> params,
+                                     Class<T> clazz, @Nullable String apiVersion)
+            throws InvalidRequestError, UnknownApiError, ServiceUnavailableError, GatewayTimeoutError, ForbiddenError,
+            RateLimitError, NotFoundError, TimeoutError, RedirectError, UnauthorizedError, MethodNotAllowedError,
+            InternalServerError, PaymentError, JsonError, HttpError, EncodingError {
```
jchen293, line 176: Same comment from above about the docstring or just use `EasyPostException`

src/test/java/com/easypost/mocking/MockClient.java
```diff
@@ -0,0 +1,217 @@
...
+    public <T> T request(Requestor.RequestMethod method, String endpoint, Map<String, Object> params, Class<T> clazz)
+            throws GatewayTimeoutError, RateLimitError, InvalidRequestError, NotFoundError, TimeoutError, EncodingError,
+            UnauthorizedError, MethodNotAllowedError, InternalServerError, UnknownApiError, ServiceUnavailableError,
+            ForbiddenError, JsonError, HttpError, RedirectError, PaymentError {
+        return processMockRequest(method, endpoint, params, clazz, null);
+    }
+
+    @Override
+    public <T> T request(Requestor.RequestMethod method, String endpoint, Map<String, Object> params, Class<T> clazz,
+                         String apiVersion)
```
jchen293, line 211: Need docstring

bnbry (commented):
I think its cool you built this and I am all for doing things like this, especially for learning purposes, but I really don't want us to get involved with supporting a mocking system like this and I also would rather take a different approach in general when it comes to solving the problems this is going after.

1) looser expectations around API calls(recorded or not) when they vary
2) extracting and directly testing helper functions with real objects instantiated with fixed data when we want precision
3) When all else fails, relying on solid open source testing tools like mockito to allow us to get around things like the billing requirements

I personally think this is rad, and I'm kind of laughing about it because I did a similar thing at a previous job(lib was called smock) but for now I want us to close out this PR and take the above approaches instead of introducing this new mocking approach.

PR #277, opened 2023-07-05
[Feature] Add library hooks

jchen293:
# Description

- Adds new `RequestHook` and `ResponseHook` classes. (un)subscribe to them with the new `subscribeToRequestHook`, `subscribeToResponseHook`, `unsubscribeFromRequestHook`, or `unsubscribeFromResponseHook` methods of an `EasyPostClient`
# Testing

Add new unit tests

# Pull Request Type


src/main/java/com/easypost/http/Requestor.java
```diff
@@ -553,7 +560,20 @@ private static <T> T httpRequest(final RequestMethod method, final String url, f
...
             }
         }
-
+        Instant requestTimestamp = Instant.now();
+        UUID requestUuid = UUID.randomUUID();
+        Map<String, String> headers = new HashMap<String, String>();
+        headers = generateHeaders(client.getApiKey());
+
+        HashMap<String, Object> requestBodyForHook = new HashMap<String, Object>();
+        requestBodyForHook.put("headers", headers);
```
nwithan8, line 569: Let's not use the Pythonic kwargs approach here for Java. We should be calling the function with six explicit parameters rather than one dictionary. Users can better account for exactly what parameters will be passed into their callback function then, instead of needing to parse a dictionary.

src/main/java/com/easypost/service/EasyPostClient.java
```diff
@@ -145,6 +156,38 @@ public EasyPostClient(String apiKey, int connectTimeoutMilliseconds, int readTim
         this.webhook = new WebhookService(this);
     }
 
+    /**
+     * Subscribes to a request hook from the given function.
+     * @param function
+     */
+    public void subscribeToRequestHook(Function<HashMap<String, Object>, Object> function) {
```
nwithan8, line 163: I feel like there was some discussion with @Justintime50 RE: whether we want to support triggering multiple callbacks in our library, or simply one pre-request and one post-request callback, which users can trigger additional callbacks from there.

nwithan8 (changes requested):

src/main/java/com/easypost/http/Requestor.java
jchen293: Having six explicit parameters will require users to have those parameters in the custom function they are passing in. If we don't like the hashmap, using an object with getters might be better

src/main/java/com/easypost/service/EasyPostClient.java
jchen293: In Python CL, which has been merged, is one callback in our lib

src/main/java/com/easypost/http/Requestor.java
nwithan8: An object could also work. I don't think the parameter order is necessarily a bad thing, would require them to know the method signature (this is already what we do with Ruby): https://github.com/EasyPost/easypost-ruby#custom-connections

src/main/java/com/easypost/service/EasyPostClient.java
nwithan8: So we don't need to subscribe to multiple callbacks?

jchen293: Not 100% sure what you mean by `multiple callbacks`

src/main/java/com/easypost/http/Requestor.java
jchen293: I think explicit parameter order is the worst approach. Users would need to know all the parameters and the correct order, if we add additional request/response values for the hooks in the future, it would be a breaking change for the users whereas using hashmap and object will not.

nwithan8: Good argument. I'd be in favor of an object then.

dependency-check-suppressions.xml
```diff
@@ -8,5 +8,6 @@
         <vulnerabilityName>CVE-2022-3171</vulnerabilityName>
         <vulnerabilityName>CVE-2022-3509</vulnerabilityName>
         <vulnerabilityName>CVE-2022-3510</vulnerabilityName>
+        <vulnerabilityName>CVE-2023-2976</vulnerabilityName>
```
nwithan8, line 11: Ref: https://nvd.nist.gov/vuln/detail/CVE-2023-2976

src/main/java/com/easypost/http/Requestor.java
```diff
@@ -553,7 +562,16 @@ private static <T> T httpRequest(final RequestMethod method, final String url, f
                     break;
             }
         }
+        Instant requestTimestamp = Instant.now();
+        UUID requestUuid = UUID.randomUUID();
+        Map<String, String> headers = new HashMap<String, String>();
+        headers = generateHeaders(client.getApiKey());
```
nwithan8, line 568: What is happening on this line?

nwithan8 (changes requested):

src/main/java/com/easypost/http/Requestor.java
jchen293: It's generating the header for the request, the function requires api key because it's part of the request header

nwithan8: nit: I think this stuff should be kicked further downstream, but I won't block on it. It's mostly just the header stuff that's not DRY here.

nwithan8 (approved):

dcaballeroc (approved):

PR #282, opened 2023-09-05
[fix] FedEx SmartPost hits register endpoint

nwithan8:
# Description

- FedEx SmartPost carrier account registration should hit `/register` endpoint for creation

# Testing

# Pull Request Type

### DIFF
--- a/src/main/java/com/easypost/Constants.java
+++ b/src/main/java/com/easypost/Constants.java
@@ -58,7 +58,7 @@ public abstract class Constants {
 
     public abstract static class CarrierAccountTypes {
         public static final List<String> CARRIER_TYPES_WITH_CUSTOM_WORKFLOW = ImmutableList.of("FedexAccount",
-                "UpsAccount");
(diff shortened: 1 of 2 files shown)


src/main/java/com/easypost/Constants.java
```diff
@@ -58,7 +58,7 @@ public abstract static class ErrorCodes {
 
     public abstract static class CarrierAccountTypes {
         public static final List<String> CARRIER_TYPES_WITH_CUSTOM_WORKFLOW = ImmutableList.of("FedexAccount",
-                "UpsAccount");
+                "UpsAccount", "FedexSmartpostAccount");
```
Justintime50, line 61: question: can these be per line or nah?

Justintime50 (approved):

src/main/java/com/easypost/Constants.java
nwithan8: Probably, but I don't feel like fighting with the Java linter right now 🙃

jchen293 (approved):

PR #286, opened 2023-09-28
[chore] Ignore vulnerability in DependencyCheck and bumped the version

nwithan8:
# Description

The DependencyCheck plugin we use to scan for vulnerabilities in our library is, itself, vulnerable. This is a red herring that unfortunately causes our CI to fail.

This PR will ignore this known vulnerability for the time being.

This does NOT pass vulnerable code down to our end-users, as this is purely an issue with one of the plugins we use to test the library.

# Testing

- `make scan` calls dependency check as expected
- Dependency scan passes as expected (no current vulnerabilities other than the aforementioned vulnerability inside the now-removed Maven plugin)

# Pull Request Type

### DIFF
--- a/dependency-check-suppressions.xml
+++ b/dependency-check-suppressions.xml
@@ -3,11 +3,16 @@
     <suppress>
         <!--
         Below vulnerabilities are from outdated Protocol Buffers which is the dependency of Error Prone.
-        This will not affect our code
+        This will not affect our code.
         -->
         <vulnerabilityName>CVE-2022-3171</vulnerabilityName>
         <vulnerabilityName>CVE-2022-3509</vulnerabilityName>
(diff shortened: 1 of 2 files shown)


dcaballeroc (dismissed):
Still failing with Github Actions though

dcaballeroc (approved):

Justintime50 (dismissed):
I don't love this particular solution. Let's chat internally about this. I feel it defeats the point of having a dependency manager if we just store local copies of the dependencies. Dependencies typically shouldn't be checked into a repo if we can help it since it's not our code and will make future diffs bigger than they should be.

dcaballeroc (approved):
Looks in line with what Justin was asking

jchen293 (approved):

PR #289, opened 2023-11-13
[fix] Address verification details lost during deserialization

### LINKED ISSUE #288: [Bug]: Address verification errors are being incorrectly parsed by the library
### Software Version

6.8.0

### Language Version

11

### Operating System

Ubuntu Linux

### What happened?

1. Create an address with verify: true, or use createAndVerify
2. API returns error objects with details of verification failures, as expected.
3. Address verification errors are being mapped incorrectly to ` {"message": "API did not return error details.", "code": "NO RESPONSE CODE"}` regardless of the actual error code and message returned by the API!

### What was expected?

Given the following API response:

`"verifications":{"zip4":{"success":false,"errors":[{"code":"E.ADDRESS.NOT_FOUND","field":"address","message":"Address not found","suggestion":null}],"details":null},"delivery":{"success":false,"errors":[{"code":"E.ADDRESS.NOT_FOUND","field":"address","message":"Address not found","suggestion":null}],"details":{}`

Expected that Error objects should have the error codes and messages received, instead of the generic fallback of "API did not return error details." and "NO RESPONSE CODE"

### Sample Code

_No response_

### Relevant logs

```sh
address.prettyPrint():

"verifications": {
    "zip4": {
      "success": false,
      "errors": [
        {
          "message": "API did not return error details.",
          "code': "NO RESPONSE CODE",
          "errors": null,
          "suggestion": null,
          "field": null
        }
      ],
      "details": null
    },
    "delivery": {
      "success": false,
      "errors": [
        {
          "message": "API did not return error details.",
          "code": "NO RESPONSE CODE",
          "errors": null,
          "suggestion": null,
          "field": null
        }
      ],
      "details": {
        "latitude": null,
        "longitude": null,
        "time_zone": null
      }
    },
```

nwithan8:
# Description

Our current setup reuses the `Error` class to represent address verification errors. However, the `Error` class has a custom deserializer that is being called during typical JSON deserialization in our library. Hard-coded logic inside that deserializer was effectively causing deserialization of the hints for address verification to fail.

This PR introduces another custom deserializer, specifically for the `AddressVerification` class that handles deserializing address verification JSON without the `Error` deserializer getting in the way.

Closes #288 

# Testing

- Improved existing unit test to verify that deserialization worked properly (that address verification details were not lost during deserialization)

# Pull Request Type

### DIFF
--- a/src/main/java/com/easypost/model/AddressVerification.java
+++ b/src/main/java/com/easypost/model/AddressVerification.java
@@ -8,4 +8,31 @@ public final class AddressVerification {
     private Boolean success;
     private List<Error> errors;
     private AddressDetail details;
+
+    /**
+     * Set the success of this address verification object.
+     *
+     * @param success The success message.
+     */
+    void setSuccess(final boolean success) {
+        this.success = success;
+    }
+
+    /**
+     * Set the errors of this address verification object.
+     *
+     * @param errors The errors.
+     */
+    void setErrors(final List<Error> errors) {
+        this.errors = errors;
+    }
+
+    /**
+     * Set the details of this address verification object.
+     *
+     * @param details The details.
+     */
+    void setDetails(final AddressDetail details) {
+        this.details = details;
+    }
 }
--- a/src/main/java/com/easypost/model/Error.java
+++ b/src/main/java/com/easypost/model/Error.java
@@ -29,4 +29,31 @@ public final class Error {
     void setCode(final String code) {
         this.code = code;
     }
+
+    /**
+     * Set the errors of this error object.
+     *
+     * @param errors The errors.
+     */
+    void setErrors(final List<Error> errors) {
+        this.errors = errors;
+    }
+
+    /**
+     * Set the suggestion of this error object.
+     *
+     * @param suggestion The suggestion.
+     */
+    void setSuggestion(final String suggestion) {
+        this.suggestion = suggestion;
+    }
+
+    /**
+     * Set the field of this error object.
+     *
+     * @param field The field.
+     */
+    void setField(final String field) {
+        this.field = field;
+    }
 }
--- a/src/test/java/com/easypost/AddressTest.java
+++ b/src/test/java/com/easypost/AddressTest.java
@@ -5,6 +5,7 @@ import com.easypost.exception.EasyPostException;
 import com.easypost.exception.General.EndOfPaginationError;
 import com.easypost.model.Address;
 import com.easypost.model.AddressCollection;
+import com.easypost.model.Error;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
@@ -14,6 +15,7 @@ import java.util.List;
 import java.util.Map;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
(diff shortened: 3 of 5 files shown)


Justintime50 (approved):
Code looks good!

src/test/java/com/easypost/AddressTest.java
```diff
@@ -79,6 +79,10 @@ public void testCreateVerify() throws EasyPostException {
         assertInstanceOf(Address.class, address);
         assertTrue(address.getId().startsWith("adr_"));
         assertEquals("417 MONTGOMERY ST FL 5", address.getStreet1());
+
+        assertNotNull(address.getVerifications());
+        assertNotNull(address.getVerifications().getZip4().getErrors()); // Should have a error due to second line
```
jchen293, line 84: This test has empty errors, so technically we are not testing if the custom deserializer (AddressVerificationDeserializer) works or not. If it's hard to reproduce the error, we could define a JSON string and write a unit test in this file like what we have done in the past: https://github.com/EasyPost/easypost-java/blob/master/src/test/java/com/easypost/ErrorTest.java

src/main/java/com/easypost/model/AddressVerification.java
```diff
@@ -8,4 +8,31 @@ public final class AddressVerification {
...
+    }
+
+    /**
+     * Set the details of this address verification object.
+     *
+     * @param details The details.
+     */
+    void setDetails(final AddressDetail details) {
+        this.details = details;
+    }
```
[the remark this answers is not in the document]
nwithan8: My concern/question about that, would that make the publicly-accessible setters or internal only?

src/test/java/com/easypost/AddressTest.java
nwithan8: Sure, I originally had a test for a specific error code, but didn't want to lock it into API-specific expectations.

src/main/java/com/easypost/model/AddressVerification.java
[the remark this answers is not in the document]
jchen293: I just found out that you can define it in Lombok by using `AccessLevel.PRIVATE`: https://projectlombok.org/features/experimental/FieldDefaults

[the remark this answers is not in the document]
nwithan8: Doesn't look like that applies to setter/getters. I'm going to keep it as is since that mimics the explicit setters we have in the `Error` class as well.

src/main/java/com/easypost/model/Error.java
```diff
@@ -29,4 +29,31 @@ void setMessage(final String message) {
...
+    }
+
+    /**
+     * Set the field of this error object.
+     *
+     * @param field The field.
+     */
+    void setField(final String field) {
+        this.field = field;
+    }
```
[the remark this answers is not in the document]
nwithan8: Keeping this since `setMessage` already existed and I just want to remain consistent.

src/test/java/com/easypost/AddressTest.java
nwithan8: Resolved

jchen293 (approved):

src/test/java/com/easypost/AddressTest.java
```diff
@@ -81,8 +83,14 @@ public void testCreateVerify() throws EasyPostException {
         assertEquals("417 MONTGOMERY ST FL 5", address.getStreet1());
 
         assertNotNull(address.getVerifications());
-        assertNotNull(address.getVerifications().getZip4().getErrors()); // Should have a error due to second line
-        assertNotNull(address.getVerifications().getDelivery().getErrors());
+
+        assertFalse(address.getVerifications().getDelivery().getErrors().isEmpty()); // should have at least one error
```
Justintime50, line 87: suggestion: the double negative is hard to read. We should probably be asserting that it is present instead

Justintime50 (approved):

src/test/java/com/easypost/AddressTest.java
nwithan8: IDE didn't like that we weren't using this "simplified" way.

PR #294, opened 2023-11-29
Change list of webhooks return type

jchen293:
# Description

This PR changes the return type for `all()` function in the webhook service from `WebhookCollection` to a list of webhooks. We made this change so it matches other type-strict client libraries (C# and Go) since this list is unpaginated.
Removed CarbonOffset model, this was missed from previous PR

# Testing

Update the unit test

# Pull Request Type


src/test/java/com/easypost/WebhookTest.java
```diff
@@ -108,12 +107,15 @@ public void testRetrieve() throws EasyPostException {
     @Test
     public void testAll() throws EasyPostException {
         vcr.setUpTest("all");
+        Map<String, Object> params = new HashMap<>();
+        params.put("url", Fixtures.webhookUrl());
 
-        WebhookCollection webhooks = vcr.client.webhook.all();
+        vcr.client.webhook.create(params);
```
Justintime50, line 113: issue: when testing the retrieval of all webhooks, we shouldn't also be creating them. Can we remove this line?

Justintime50 (changes requested):

src/test/java/com/easypost/WebhookTest.java
jchen293: It would be an empty list of webhooks. We need to create one to ensure the webhook objects deserialize into a list properly

src/main/java/com/easypost/Constants.java
```diff
@@ -75,7 +77,8 @@ public abstract static class Http {
                 .registerTypeAdapter(SmartrateCollection.class, new SmartrateCollectionDeserializer())
                 .registerTypeAdapter(Error.class, new ErrorDeserializer())
                 .registerTypeAdapter(AddressVerification.class, new AddressVerificationDeserializer())
-                .registerTypeAdapter(StatelessRate[].class, new StatelessRateDeserializer()).create();
+                .registerTypeAdapter(StatelessRate[].class, new StatelessRateDeserializer())
```
nwithan8, line 80: These type adapters are starting to get out of hand...

nwithan8 (approved):

src/test/java/com/easypost/WebhookTest.java
Justintime50: I suppose that's true eh... thanks for clarifying

Justintime50 (approved):

PR #297, opened 2023-12-07
add missing object field to all models

jchen293:
# Description

Add the missing `object` field to all models. Previously, some models had the `object` field while most models did not, moving the `object` field to `EasyPostSource.java` so every model has this field

# Testing

Re-record a dozen cassettes, add an assertion to one unit test

# Pull Request Type


Justintime50 (commented):
My only concern is does this no pop up on object that do NOT have an object and we essentially flipped the problem?

jchen293:
> My only concern is does this no pop up on object that do NOT have an object and we essentially flipped the problem?

If a user tries to get the `object` field when the endpoint does not have it, then it returns a `null`. The concept here is the same as having `id`, `created_at`, and `updated_at` fields in `EasyPostSource.java` for all models as almost all of our endpoints return an `object` field as part of the response, if a model doesn't have an `object` field, we can explicitly define the fields for that model class and not inherit to the `EasyPostSource.java`

Justintime50 (approved):
What we have seems reasonable

PR #311, opened 2024-04-09
[fix] ID field usage in string representation of EasyPost objects

nwithan8:
# Description

Previous method of reflection to retrieve the `id` of an `EasyPostResource` often failed, resulting in fallback to empty string. As a result, the string representation of any `EasyPostResource` was often an empty string.

New method improves `toString` calculation, including accounting for present/non-present `id` field (ephemeral objects).

# Testing

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,9 @@
 
 ## Next Release
 
-- Fix payment method funding and deletion failures due to undetermined payment method type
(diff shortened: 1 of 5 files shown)


jchen293:
I know the code is very old, but is it possible to have a unit test for this change?

jchen293 (approved):

Justintime50 (approved):

PR #312, opened 2024-04-12
[fix] Invalid `Fields` structure

nwithan8:
# Description

Port of [fix in .NET library](https://github.com/EasyPost/easypost-csharp/pull/558) to Java.

Fun fact: The `Fields` object in this library previously did not have any properties declared. This means there was never any deserialization issues, but also means that each `Field` was never reachable by end-users.

# Testing

- Add unit test to confirm deserialization works as expected
  - The carrier accounts used in our existing unit tests do not have fields returned by the API, so the serialization issue was not previously caught.
  - Beyond that, the properties of `Fields` were not declared, so they were never deserialized.

# Pull Request Type

### DIFF
--- a/src/test/java/com/easypost/BillingTest.java
+++ b/src/test/java/com/easypost/BillingTest.java
@@ -25,15 +25,6 @@ public final class BillingTest {
             "ll,\"last4\":\"4444\",\"exp_month\":1,\"exp_year\":2025,\"brand\":\"Mastercard\"}}";
     private PaymentMethod paymentMethod = Constants.Http.GSON.fromJson(jsonResponse, PaymentMethod.class);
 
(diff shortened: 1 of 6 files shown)


src/test/java/com/easypost/BillingTest.java
```diff
@@ -25,15 +25,6 @@ public final class BillingTest {
             "ll,\"last4\":\"4444\",\"exp_month\":1,\"exp_year\":2025,\"brand\":\"Mastercard\"}}";
     private PaymentMethod paymentMethod = Constants.Http.GSON.fromJson(jsonResponse, PaymentMethod.class);
 
-    private String jsonResponseLegacyPrefixes = "{\"id\":\"cust_...\",\"object\":\"PaymentMethods\",\"primary_" +
```
jchen293, line 28: nice catch

jchen293 (approved):

Justintime50 (approved):

